PR #7, opened 2023-05-23
Initial C# Implementation

MarioGK:
This is the current lexer implementation for C#, the best language in the world

Added a docker file and a .sh to make it easier to run as well


ThePrimeagen:
TLDR

i am sorry for you
congratulations

Jvp2001:
@MarioGK What about using spans instead of strings? This would help with performance and memory allocations.

MarioGK:
@Jvp2001 Hello, yeah i plan to replace all string manipulation with span, but just to get things working in a faster and simpler way i went the normal string route for now

PR #10, opened 2023-05-25
Big P(H)P language

wartab:
Haven't written any PHP code in 10 years, wanted to check if it's bad. It's better than before, but still sucks. 

Shoutouts to half assed implementation of pattern matching, forcing you to basically use it in a hacky way to be able to use it.

Also didn't bother doing Tests the fancy way with PHPUnit, I have completely forgotten how to use it, but it's got a nice test.php


ThePrimeagen:
literally only looked to see if it has its own folder

check

please continue implementing :)

PR #12, opened 2023-05-25
Java lexer

xNaCly:
I know the language is slow, but I guess it's important to have diverse language benchmarks


ThePrimeagen:
Ok, we have two Java PRs

can we please figure out which one is correct? https://github.com/ThePrimeagen/ts-rust-zig-deez/pull/13

PR #13, opened 2023-05-25
Java lexer built with Gradle

PossessedC0bra:
Hungarian notation goes brrrr


ThePrimeagen:
Hey, there is already a PR for java

could you two, @xNaCly figure out which to use?  #12

xNaCly:
Well mine is more minimal and i was first, but if you prefer to use gradle and all the stuff around it sure. The implementation seems to be around the same, just with more oop on @PossessedC0bra part.

PossessedC0bra:
@ThePrimeagen use @xNaCly implementation. His solution is more minimal. I made my solution bloatet on purpose to poke fun at some of your funny comments against such bloat. The amount of bloat is mostly just a shitpost :)

PossessedC0bra:
Closed in favor of java implementation found in #12

PR #16, opened 2023-05-25
Lexer implementation in C++

nicdgonzalez:
Compiled using Clang++ 16.0.1

I'm still trying to figure out how to use Docker (uh oh) so I will add in the Dockerfile once I can get that running as requested in one of the other PRs.

Requires C++23 but can easily be updated to run on C++17 by replacing `std::to_underlying` to `static_cast<int>` in `tests.cc`. I was waiting to see if Docker supported C++23 before making the change since `to_underlying` is better than hard coding the type.


ThePrimeagen:
looks memory safe to me

cpp/lexer/lexer.cc
```diff
@@ -0,0 +1,92 @@
+#include "lexer.hh"
+
+
+#include <exception>
+
+
+lexer::lexer(const std::string input)
+  : input_{input},
```
codebrainz, line 8: Wouldn't it make more sense to pass the `input` argument either by `const&` reference or non-`const` value then `std::move()`-ing it into `input_`? It seems like passing by `const` value will guarantee to create an extra copy of the input string, unless I'm mistaken.

I haven't read the book this is based on (yet), but assuming the input string outlives the lexer, it could also use `std::string_view` to avoid any copies at all.

Anyway, just happened here after watching a Primeagen video about this project and was glad to see someone representing C++, nice work! :+1:

PR #26, opened 2023-05-26
Go deez

hpiaia:


go/lexer.go
```diff
@@ -0,0 +1,126 @@
...
+
+	if unicode.IsDigit(l.ch) {
+		return NewToken(INT, l.readNumber())
+	}
+
+	return NewToken(ILLEGAL, string(l.ch))
+}
+
+func getTokenFromMap(ch rune) (Token, bool) {
+	var tokenMap = map[rune]Token{
```
nieomylnieja, line 81: This can and should be declared as a pkg scoped variable, you don't want to allocate memory for this map each time you call `getTokenFromMap`, since you don't modify it and only use it for mapping.

After you do that, the function will become obsolete too.

nieomylnieja (changes requested):

go/lexer.go
hpiaia: Good catch, fixed.

PR #34, opened 2023-05-26
Minimal implementation using Lazarus with Free Pascal

attilakapostyak:
A minimal implementation in Pascal, just for fun.

Dev Environment: Lazarus (Free Pascal)

### DIFF
--- a/pascal/Lexer.lpi
+++ b/pascal/Lexer.lpi
@@ -1,0 +1,64 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<CONFIG>
+  <ProjectOptions>
+    <Version Value="12"/>
+    <PathDelim Value="\"/>
+    <General>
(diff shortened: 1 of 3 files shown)


zivlakmilos:
Why not use just free pascal compiler? That way You can create Docker for building without relay on specific IDE like Lazarus

attilakapostyak:
I started it to have Pascal as well, but I agree with you, in the end FreePascal is the better option. However, I opted for now to do it quickly in Windows Sandbox with the IDE. When I have some time I will convert it to FreePascal and also have the Dockerfile for building.

novusnota:

ThePrimeagen (approved):

ThePrimeagen:
adding you to code owners @attilakapostyak and please add a docker image to run this.

PR #36, opened 2023-05-26
refactor(fsharp): use readWhile for reading multiple characters

meenzen:
Inspired by @tjdevries OCaml implementation


tjdevries:
i'll allow it

PR #40, opened 2023-05-26
Jai

marvhus:
I decided to make a Jai version.

I'm keeping it as a draft for now since I haven't tested the docker stuff for it.

Note:
The Makefile does not include any compiling since Jai is stil in private beta, but I have included an elf binary that should work. I might compile one for windows too.

Note 2:
This is just a simple implementation straight from the book, and I have only used Jai for a few weeks, so I'm still not entirely sure what optimizations I can do, but it works, and it runs the test in 0.008s ... so I'm happy.

For anyone curios, this is the output when I compile:
image


ThePrimeagen:
i am getting the go ahead from Jonathan before i merge this in.

give me a moment pls :)

marvhus:
@ThePrimeagen Any updates on if you'll add this?

ThePrimeagen:
beautiful and sorry

PR #42, opened 2023-05-26
add clojure implementation

cvanloo:
I'm very new to Clojure, so this probably isn't the most "idiomatic" solution -- thought I would give it a go anyway.

Any suggestions/criticism are welcome.

### DIFF
--- a/clojure/Dockerfile
+++ b/clojure/Dockerfile
@@ -1,0 +1,12 @@
+FROM clojure:latest
+
+RUN mkdir /deez
+WORKDIR /deez
+COPY . .
+# @fixme: this is somehow missing a few deps!
+RUN clj -A:deps -A:cljfmt -A:test -A:lint
+#RUN clj -X:deps prep
+
+CMD make fmt && \
+    make lint && \
+    make test
--- a/clojure/src/monkey.clj
+++ b/clojure/src/monkey.clj
@@ -1,65 +1,61 @@
 (ns monkey
   (:import [java.lang Character]))
 
-(def keywords
-  {"fn" :function
-   "let" :let})
-
-(defn monkey-keyword?
-  [word]
-  (contains? keywords word))
+(def kind first)
+(def position second)
+(def literal #(nth % 2 nil))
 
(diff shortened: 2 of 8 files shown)


clojure/src/monkey.clj
```diff
@@ -0,0 +1,77 @@
...
+                       (= \( current) (make-token :lparen)
+                       (= \) current) (make-token :rparen)
+                       (= \, current) (make-token :comma)
+                       (= \; current) (make-token :semicolon)
+                       (= \+ current) (make-token :plus)
+                       (= \= current) (make-token :equal)
+                       (letter? current) (read-word-or-token input)
+                       (monkey-number? current) (read-number input)
+                       :else {:token :illegal, :literal current})]
+       (recur (skip-whitespace (drop n input))
```
cvanloo, line 76: I'm not sure if the way `n` is calculated is a good idea, but it was the simplest to implement right away.

Vikasg7:
@cvanloo I have added a Clojure implementation, you can check it out. it's very short and concise.

clojure/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM clojure:latest
+
+RUN mkdir /deez
+WORKDIR /deez
+COPY . .
+# @fixme: this is somehow missing a few deps!
+RUN clj -A:deps -A:cljfmt -A:test -A:lint
+#RUN clj -X:deps prep
```
cvanloo, line 8: I'm trying to cache the deps here, so that they don't need to be downloaded on every `docker run`.
However, this seems to miss a few? The commented-out command below also didn't work.

cvanloo:
> @cvanloo I have added a Clojure implementation, you can check it out. it's very short and concise.

Cool! Your implementation looks really clean. I didn't even know about the `case` function. Making the lexer a lazy-seq I found also impressing.

I'm closing this PR in favor of @Vikasg7 solution.

cvanloo:
I've cleaned my solution up a bit, leaving this here for reference.

PR #43, opened 2023-05-26
make Roc implementation way simpler and more idiomatic

bhansconnect:
The first version was my personal version that was messing around with Data Oriented Design and other hacks.
This version is way cleaner, easier to follow, and more idiomatic.
Should be much closer, to other functional languages.

### DIFF
--- a/roc/src/Eval.roc
+++ b/roc/src/Eval.roc
@@ -1,13 +1,18 @@
 interface Eval
     exposes [eval, evalWithEnv, newEnv, printValue]
-    imports [Lexer, Parser.{ Index, Node, ParsedData }]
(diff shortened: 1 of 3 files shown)


bhansconnect (commented):
The only struggle with the CodeOwners setup is that I can't approve my own pull request even if it only touches roc code.

Anyway, this is tested, just updates Roc code, and gets rid of some overly complex ways of doing things. Instead it is much simpler and should be more readable to most people.

ThePrimeagen:
will be merging :)

ThePrimeagen (approved):

ThePrimeagen:
real talk, i have NO idea what this does

@bhansconnect i am going to make you code owner and i want you to review any more roc prs :)

PR #46, opened 2023-05-26
feat(cobol): COBOL deez nuts (only single char tokens)

kkard2:
i should be sleeeping rn instead of doing this ;_;

i really shouldn't do it but i probably will be tempted to continue implementing the full lexer 😔


novusnota:

ThePrimeagen:
Please make sure you're on the code owners for Cobol

PR #49, opened 2023-05-27
`elixir` Rewrite it now that I have time

ryanwinchester:
### Changed

* _Everything_

### Removed

* Token struct removed in favour of atoms and tuples

### Added

* Typespecs
* Tests
* Dockerfile

Basically, all the Lexer code is in [this module](https://github.com/ThePrimeagen/ts-rust-zig-deez/blob/27e3310d19f15b65bab6dd0a84695dc35d518de4/elixir/lib/lexer.ex) now.

## Notes for those new to Elixir

The typespecs are not necessary but make it nicer for the user with LSP and static analysis/type-checking tools.
Screenshot 2023-05-29 at 7 30 40 PM

The examples in the `@doc` comments are run as part of the test suite.

```elixir
  @doc """
  Turn some input into a list of tokens.

  ## Example

      iex> Lexer.init("let five = 5;")
      [
        :let,
        {:ident, "five"},
        :assign,
        {:int, "5"},
        :semicolon,
        :eof
      ]

  """
```
Screenshot 2023-05-29 at 7 34 42 PM

The `@moduledoc`, `@typedoc`, and `@doc` comments get parsed as markdown and display nicely in HTML or in the interactive shell. e.g. `h Monkey.Lexer.init` (`h` for `help`)

Screenshot 2023-05-29 at 7 37 36 PM


elixir-shyryan/lib/lexer.ex
```diff
@@ -1,77 +1,138 @@
...
+  defp tokenize(<<"=", rest::binary>>), do: {:assign, rest}
+  defp tokenize(<<"+", rest::binary>>), do: {:plus, rest}
+  defp tokenize(<<"-", rest::binary>>), do: {:minus, rest}
+  defp tokenize(<<"!", rest::binary>>), do: {:bang, rest}
+  defp tokenize(<<"/", rest::binary>>), do: {:slash, rest}
+  defp tokenize(<<"*", rest::binary>>), do: {:asterisk, rest}
+  defp tokenize(<<">", rest::binary>>), do: {:greater_than, rest}
+  defp tokenize(<<"<", rest::binary>>), do: {:less_than, rest}
+  defp tokenize(<<"return", rest::binary>>), do: {:return, rest}
+  defp tokenize(<<c::8, rest::binary>>), do: {{:illegal, <<c>>}, rest}
```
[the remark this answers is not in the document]
ryanwinchester: @charlieroth _Always_ use pattern-matching when you can. So `case` would be acceptable here, but `cond` would not be.

You could use this as either a `case` condition or functions. I think the function approach is preferred by more, but I do know people who prefer `case`, so I wouldn't reject a PR that had `case` instead.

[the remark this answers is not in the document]
charlieroth: Okay understood, ty

ryanwinchester:
> I will bow out of this race, I know when I have been defeated 🙂

I'm not trying to compete 😅 I just want Prime to get a proper impression of idiomatic Elixir since he already doesn't view it in a very positive light. I've been using it as my predominant language full-time for like 6 years so I have a bit of experience with it is all.

elixir-shyryan/lib/lexer.ex
```diff
@@ -1,77 +1,143 @@
...
-      |> Token.lookup_identifier()
-      |> Token.new()
+  # Recusrively read the input until we hit a non-letter character. Then
+  # tokenize the word.
+  @spec read_identifier(String.t(), String.t()) :: {token(), String.t()}
+  defp read_identifier(input, acc \\ "")
 
-    tokenize(rest, [token | tokens])
+  defp read_identifier(<<c::8, rest::binary>>, acc) when is_letter(c) do
+    read_identifier(rest, acc <> <<c>>)
```
madlep, line 114: Would probably be more performant to accumulate as an iodata list. Benchmarked building a string a character at at time, and runs ~1.7x faster. If the identifier gets huge (because of pathological data), it uses 60% less memory, and runs about the same speed as `<>` concatenation.

```suggestion
    read_identifier(rest, [acc | <<c>>])
```

elixir-shyryan/lib/lexer.ex
```diff
@@ -1,77 +1,143 @@
...
   end
 
-  defp accumulate_identifier(<<c::8, rest::binary>>, acc) when is_letter(c) do
-    accumulate_identifier(rest, acc <> <<c>>)
+  defp read_identifier(rest, identifier) do
+    {tokenize_word(identifier), rest}
   end
```
madlep, line 119: Turn the iodata value into a binary

```suggestion
  defp read_identifier(rest, acc) do
    identifier = 
      acc
      |> IO.iodata_to_binary()
      |> tokenize_word()
    
    {identifier, rest}
  end
```

elixir-shyryan/lib/lexer.ex
```diff
@@ -1,77 +1,143 @@
...
+  # Recursively read the input until we hit a non-digit character. Then
+  # tokenize the number.
+  @spec read_number(String.t(), String.t()) :: {token(), String.t()}
+  defp read_number(input, acc \\ "")
 
-  defp read_number(input, tokens) do
-    {number, rest} = accumulate_number(input, "")
-    tokenize(rest, [Token.new(:int) | tokens])
+  defp read_number(<<c::8, rest::binary>>, acc) when is_digit(c) do
+    read_number(rest, acc <> <<c>>)
```
madlep, line 127: Could do same thing accumulating using `IO.iodata_to_binary/1` here too.

madlep (approved):
Looks awesome 👍

elixir-shyryan/lib/lexer.ex
```diff
@@ -1,77 +1,143 @@
...
+  defp tokenize(<<"}", rest::binary>>), do: {:rsquirly, rest}
+  defp tokenize(<<"=", rest::binary>>), do: {:assign, rest}
+  defp tokenize(<<"+", rest::binary>>), do: {:plus, rest}
+  defp tokenize(<<"-", rest::binary>>), do: {:minus, rest}
+  defp tokenize(<<"!", rest::binary>>), do: {:bang, rest}
+  defp tokenize(<<"/", rest::binary>>), do: {:slash, rest}
+  defp tokenize(<<"*", rest::binary>>), do: {:asterisk, rest}
+  defp tokenize(<<">", rest::binary>>), do: {:greater_than, rest}
+  defp tokenize(<<"<", rest::binary>>), do: {:less_than, rest}
+  defp tokenize(<<"return", rest::binary>>), do: {:return, rest}
```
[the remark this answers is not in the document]
ryanwinchester: Yeah, you're right, and then gets handled in `tokenize_word`.

elixir-shyryan/lib/lexer.ex
madlep: With iodata lists you can. Allows for "improper" lists that can be built up in arbitrary order by wrapping another layer around it. Which works in constant time.

`IO.iodata_to_binary/1` then figures out how to pull it apart and stitch it together. Is best for appending longer strings, but works nicely a character at a time too. It's how Phoenix renders templates out to a TCP socket under the hood to avoid lots of string concatenation stitching together static template text and dynamic data.

```elixir
iex(1)> acc = ""
""
iex(2)> acc = [acc | "add"]
["" | "add"]
iex(3)> acc = [acc | [" ", "some"]]
[["" | "add"], " ", "some"]
iex(4)> acc = [acc | " text"]
[[["" | "add"], " ", "some"] | " text"]
iex(5)> acc |> IO.iodata_to_binary
"add some text"
```

ryanwinchester: if you are prepending `acc` does it not get copied every time?

madlep: It's just wrapping `acc` in a new list with an extra element. Runs in constant time, as it's leaving `acc` where it is, and just adding a reference to it in the new iodata list.

So rather than treating it as a cons list where a new node is prepended to an existing linked list, it's more like a tree where a new tree is created at each step that wraps the original tree, and can create a new list with elements before and/or after it. That runs in $O(1)$ time for each step.

Then `IO.iodata_to_binary/1` runs in `$O(n)$ time at the end to effectively do an in-order tree traversal, building an binary in one pass.

ryanwinchester: Very nice.

```elixir
alias Monkey.Lexer
alias Monkey.LexerIOData

sm = """
if (5 < 10) {
  return true;
} else {
  return false;
}
"""

md = String.duplicate(sm, 100)

lg = String.duplicate(sm, 10_000)

xl = String.duplicate(sm, 1_000_000)

Benchee.run(
  %{
    "iodata" => &LexerIOData.init/1,
    "concat" => &Lexer.init/1
  },
  inputs: %{
    "Small" => sm,
    "Medium" => md,
    "Large" => lg,
    "X-Large" => xl
  },
  memory_time: 2,
  reduction_time: 2
)
```

```
❯ mix run bench/string_builder.exs
Operating System: macOS
CPU Information: Intel(R) Core(TM) i5-9600K CPU @ 3.70GHz
Number of Available Cores: 6
Available memory: 32 GB
Elixir 1.14.3
Erlang 25.2.3

Benchmark suite executing with the following configuration:
warmup: 2 s
time: 5 s
memory time: 2 s
reduction time: 2 s
parallel: 1
inputs: Large, Medium, Small, X-Large
Estimated total run time: 1.47 min

Benchmarking concat with input Large ...
Benchmarking concat with input Medium ...
Benchmarking concat with input Small ...
Benchmarking concat with input X-Large ...
Benchmarking iodata with input Large ...
Benchmarking iodata with input Medium ...
Benchmarking iodata with input Small ...
Benchmarking iodata with input X-Large ...

##### With input Large #####
Name             ips        average  deviation         median         99th %
iodata         32.76       30.53 ms     ±2.47%       30.48 ms       33.80 ms
concat         20.77       48.14 ms     ±2.58%       47.80 ms       53.36 ms

Comparison:
iodata         32.76
concat         20.77 - 1.58x slower +17.61 ms

Memory usage statistics:

Name      Memory usage
iodata        24.31 MB
concat        24.68 MB - 1.02x memory usage +0.37 MB

**All measurements for memory usage were the same**

Reduction count statistics:

Name           average  deviation         median         99th %
iodata        972.31 K     ±0.16%       972.10 K       979.36 K
concat        938.20 K     ±0.38%       937.53 K       952.46 K

Comparison:
iodata        972.10 K
concat        938.20 K - 0.96x reduction count -34.10864 K

##### With input Medium #####
Name             ips        average  deviation         median         99th %
iodata        6.07 K      164.86 μs    ±37.11%      158.01 μs      332.60 μs
concat        4.69 K      213.08 μs    ±32.52%      201.05 μs      403.18 μs

Comparison:
iodata        6.07 K
concat        4.69 K - 1.29x slower +48.22 μs

Memory usage statistics:

Name      Memory usage
iodata       238.67 KB
concat       255.50 KB - 1.07x memory usage +16.83 KB

**All measurements for memory usage were the same**

Reduction count statistics:

Name   Reduction count
iodata         10.09 K
concat          9.85 K - 0.98x reduction count -0.24700 K

**All measurements for reduction count were the same**

##### With input Small #####
Name             ips        average  deviation         median         99th %
iodata      295.15 K        3.39 μs  ±2128.96%        0.99 μs        3.01 μs
concat      267.66 K        3.74 μs  ±2043.47%        1.10 μs        3.76 μs

Comparison:
iodata      295.15 K
concat      267.66 K - 1.10x slower +0.35 μs

Memory usage statistics:

Name      Memory usage
iodata         2.85 KB
concat         2.80 KB - 0.98x memory usage -0.04688 KB

**All measurements for memory usage were the same**

Reduction count statistics:

Name   Reduction count
iodata              89
concat              84 - 0.94x reduction count -5

**All measurements for reduction count were the same**

##### With input X-Large #####
Name             ips        average  deviation         median         99th %
iodata         0.147         6.79 s     ±0.00%         6.79 s         6.79 s
concat        0.0855        11.69 s     ±0.00%        11.69 s        11.69 s

Comparison:
iodata         0.147
concat        0.0855 - 1.72x slower +4.91 s

Memory usage statistics:

Name      Memory usage
iodata         2.42 GB
concat         2.47 GB - 1.02x memory usage +0.0468 GB

**All measurements for memory usage were the same**

Reduction count statistics:

Name   Reduction count
iodata        103.20 M
concat        114.69 M - 1.11x reduction count +11.49 M

**All measurements for reduction count were the same**
```

I learned something today.  I've never used an improper list like this. I think because my LSP complains at me if I end up with one.

elixir-shyryan/lib/lexer.ex
```diff
@@ -1,77 +1,143 @@
...
+  # Tail-recursively go through the input, tokenizing the character(s).
+  defp lex(input, tokens) do
+    {token, rest} = tokenize(input)
+    lex(rest, [token | tokens])
+  end
+
+  # Uses binary pattern-matching to match on the first character(s).
+  # See: https://hexdocs.pm/elixir/Kernel.SpecialForms.html#<<>>/1
+  @spec tokenize(String.t()) :: {token(), String.t()}
+  defp tokenize(<<c::8, _::binary>> = input) when is_letter(c), do: read_identifier(input)
```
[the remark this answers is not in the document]
ryanwinchester: I'm fine with ascii support only for now unless I get some requirements that specify utf8

elixir-shyryan/lib/lexer.ex
madlep: Awesome. Glad it helped 😀 . It's a handy trick. I think I've only ever used it a handful of times, but it's useful when it fits.

Performance difference is even larger when there are long strings being put together. And the fact that you can nest arbitrarily deep, and add an arbitrary number of elements before/after at each step.

Interesting that dialyzer complains about that. Would have thought it would understand `iodata()` type - as that's it's exact use case.

elixir-shyryan/lib/lexer.ex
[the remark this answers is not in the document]
ryanwinchester: re: `when c in ?a..?z or c in ?A..?Z or c == ?_`

elixir-shyryan/lib/lexer.ex
madlep: Seems like dialyzer warns about it because it's _usually_ not what you'd want to do, and probably a mistake in most cases.

If you pass an improper list to most functions expecting a regular cons list, it'll error.

Advice seems to be to add an ignore attribute to dialyzer to silence it 🤷‍♂️ https://github.com/erlang/otp/issues/5937

ryanwinchester: Seems like a not great default for dialyzer there.

sbotzek:
Note that there's currently a bug in this: it won't let identifiers start with "return", "let", etc.  Here's a test I tried running:
`
    test "literals can start with return" do
      input = """ 
      returnfds
      """

      expected_tokens = [ 
        {:ident, "returnfds"},
        :eof
      ]   

      assert Lexer.init(input) == expected_tokens      
`

Results in:
`
  1) test lex/1 literals can start with return (LexerTest)
     test/lexer_test.exs:207
     Assertion with == failed
     code:  assert Lexer.init(input) == expected_tokens
     left:  [:return, {:ident, "fds"}, :eof]
     right: [{:ident, "returnfds"}, :eof]
     stacktrace:
       test/lexer_test.exs:217: (test)
`

sbotzek:
Note that there's a bug in this - it won't let identifiers start with "return", "let", etc.  To see the bug manifest, try to tokenize "returnfds".

ryanwinchester:
> Note that there's a bug in this - it won't let identifiers start with "return", "let", etc. To see the bug manifest, try to tokenize "returnfds".

Nice catch. Updated and added test.

ryanwinchester:
### LITHIUM-APPROVED, LET'S MERGE

Screenshot 2023-05-30 at 7 09 38 PM

image

vhladko:
Hey guys, so I'll merge this one, as I understood @charlieroth is fine with that, you can add more improvements in future PR's :)

PR #51, opened 2023-05-27
feat(haskell): added haskek implementation

alexjercan:
because someone had to do it


evan-hines-js:
It's gorgeous. 💯

Ofenhed:
Wow, you really went all in with showing multiple methods. Very nice!

alexjercan:
> Wow, you really went all in with showing multiple methods. Very nice!

Thanks. I also got some help from fellow haskelers :)

novusnota:

alexjercan:
> Merging in https://github.com/novusnota/monkey-deez

Hi, I added one more implementation from one PR. 

@newmaidumosa can you look over it? I haven't used lens before, but I think this is how you would do it. Uhm also probably need to change the readme.

newmaidumosa:
Yeah, looks good.  I thought that it would be nice to show different options and you have done so perfectly!

Disclaimer: I am not an expert, so my code might not be the best reference.

haskell/README.md
```diff
@@ -63,26 +109,26 @@ with lenses we can do this
...
+Haskell `String` type is actually a linked list of `char`. This is not very
+performant, e.g., `length` and indexing `(!!)` operations are O(N). Also,
+indexing is unsafe (can fail) with crafting or importing safe operations. With
+`ByteString` we have O(1) operations and built in safe indexing that return
+`Maybe` types (Option in Rust).
 
 ## [Control.Monad.Trans.State](https://hackage.haskell.org/package/transformers-0.6.1.0/docs/Control-Monad-Trans-State.html)
 
-since we are working with immutable variables, we have to keep passing the state around. You can see this in the `OCaml` version where `lexer` is being parsed around recursively and returned in a tuple?.
+since we are working with immutable variables, we have to keep passing the
```
Ofenhed, line 125: I would say that this is because our functions are pure, and not that the variables are immutable. Since we can't have side effects, we need to pass along the variables.

The immutability makes it so that we keep having to create new data besides our old data, and make sure that we use the right variable name.

haskell/README.md
```diff
@@ -63,26 +109,26 @@ with lenses we can do this
...
+state around. You can see this in the `OCaml` version where `lexer` is being
+parsed around recursively and returned in a tuple?.
 
-The `state` monad allows us to obfuscate the passing around of the state and perform updates on the state within a function without the need for new variable names.
+The `state` monad allows us to obfuscate the passing around of the state and
+perform updates on the state within a function without the need for new
+variable names.
```
Ofenhed, line 131: And the risk of accidentally use of old data that is no longer valid.

Ofenhed (commented):
I've got no power here, feel free to ignore these, but here are a few things in the README that caught my eye.

haskell/README.md
```diff
@@ -1,16 +1,62 @@
...
+`Lexer.Lens` is an implementation of a lexer using State and Lens.
+
+## Why use certain packages?
+
+### [Control.Lens](https://hackage.haskell.org/package/lens-5.2.2/docs/Control-Lens.html)
+
+Record syntax in Haskell is not very user friendly. Reading and updating fields
+is verbose and using lenses makes it more like to to other languages where
+`recored^.field` is equivalent to `record.field` and `record & field .~ value`
+is equivalent to `record.field = value`
```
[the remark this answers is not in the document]
alexjercan: Yeah, that looks better.

haskell/README.md
```diff
@@ -107,7 +153,13 @@ getToken = do              -- no need to manaully track the state & value of the
     _   -> modify readcahr >> pure (Token EOF)
 ```
 
-## External - Make.cmd
+## Some weird things
+
+1. How to deal with `dist-newstyle` when building with docker? Since the
+   Makefile mounts the `$pwd` to the container it generates garbage files in
+   the host filesystem. My workaround was to run `cabal clean` at the end, but
+   is there a better way?
```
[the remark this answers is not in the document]
alexjercan: I would have liked to leave the Makefile and Dockerfile as similar to the ones from the template. I would personally do a `COPY . .` since that works with a ignore file. I think the point of that mount is to be able to run the formatter on the code. But maybe there is also some other reason? @novusnota

haskell/README.md
alexjercan: Yes, this can be rephrased. Having the lexer as both input and output is basically keeping state. Maybe show an example like `String -> (a, String)` could be rewritten as a `State String a`.

haskell/README.md
```diff
@@ -63,26 +109,26 @@ with lenses we can do this
...
 
 ## [Control.Monad.Trans.State](https://hackage.haskell.org/package/transformers-0.6.1.0/docs/Control-Monad-Trans-State.html)
 
-since we are working with immutable variables, we have to keep passing the state around. You can see this in the `OCaml` version where `lexer` is being parsed around recursively and returned in a tuple?.
+since we are working with immutable variables, we have to keep passing the
+state around. You can see this in the `OCaml` version where `lexer` is being
+parsed around recursively and returned in a tuple?.
 
-The `state` monad allows us to obfuscate the passing around of the state and perform updates on the state within a function without the need for new variable names.
+The `state` monad allows us to obfuscate the passing around of the state and
```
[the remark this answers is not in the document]
alexjercan: hide is a good word I think. I wanted to quote "obfuscate" when I merged the readme, but at the same time I didn't want to gatekeep :laughing:

haskell/README.md
alexjercan: For me it looks like it makes sense, but for a beginner it might make it confusing since they would expect that State is a function itself. But it actually needs runState to run it.

I think that it might make more sense to be able to view it as

```rust
fn next_token(lexer: Lexer) -> (Token, Lexer) {
    // Logic
}
```

would be written as

```rust
impl Lexer {
    fn next_token(&mut self) -> Token {
        // Logic
    }
}
```

but this isn't really true either, because State doesn't really mutate anything

haskell/README.md
```diff
@@ -63,51 +109,63 @@ with lenses we can do this
...
+passing the state around. You can see this in the `Lexer.Basic` version where
+the lexer (a simple `String` in that case) is being parsed around recursively
+and returned in a tuple `nextToken :: Lexer -> (Token, Lexer)`. In languages
+like Rust we can implement a function for a structure to mutate it `pub fn
+next_token(&mut self) -> Token`.
 
-The `state` monad allows us to obfuscate the passing around of the state and perform updates on the state within a function without the need for new variable names.
+The `state` monad allows us to obfuscate the passing around of the state and
+perform updates on the state within a function without the need for new
+variable names.
```
Ofenhed, line 134: I would still add that it allows us to simulate mutability without actually having mutability. "Passing around of the state" implies that we have a state that we pass around, when in "reality" (ignoring compiler optimizations) we create a new state for every function call.

haskell/README.md
Ofenhed: True, but I still prefer your explanation. Even if it's not equivalent, it's the same function written as the best practice for the different languages, and it makes it clear what functions the parts of the type definition has.

haskell/README.md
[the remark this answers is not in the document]
novusnota: So, there are multiple points of my template of a Dockerfile as I see them:
- To make it easily debuggable: there's no need to peak inside the built image for you guys, and you could simply focus on correct RUN commands
- To make it super easy for anyone not involved in a development of a particular implementation to build and see all the outputted files in the build. That makes everyone being able to mess around even if you don't know the language at all — great for learning by trial-and-error, and seeing how it works

And to deal with the junk from the position of maintainers there are three things:
- .gitignore
- `Makefile` with a `make clean` target 
- (unmerged, but was in plans) `CMD` section of Dockerfile, which would actually be `make clean && make ready`.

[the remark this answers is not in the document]
novusnota: Oh, and of course you can add more Dockerfiles as you see fit, say a `impl.Dockerfile`, which would suit your needs the best as a maintainer 👍

haskell/README.md
alexjercan: I see your point. I will try to find a better explanation for this one.

vhladko:
@alexjercan thanks, I’ll merge this one and you can continue work on improvements in a future PRs :)

PR #55, opened 2023-05-27
Kotlin implementation

waymirec:


waymirec:
there are already a handful of these implementations in review so i'm going to pull mine. no need for another one.

PR #57, opened 2023-05-27
Dart: Use more language features

shreyassanthu77:
Updated the code to the following dart features:
- Pattern Matching
- Iterators
- Extension methods ( kinda like Traits in rust )


marelle65187512:
I noticed a bug in your code, here's a possible fix:

shreyassanthu77:
WHAT IS THE BUG!!

JikoUnderscore:
`String c when c.isLetter() => _lookupIdent(
          _readIdentifier(),
        ),
      String c when c.isNumber() => Token(
          TokenType.int,
          _readNumber(),
        ),`

These switch expirations should return early and never call `_readChar();`.
I think this is the bug that @marelle65187512 referred too.

shreyassanthu77:
Yes but I do it a bit differently. I go back after calling read* function so that should not be an issue

D-James-GH:
If using the new dart language features could we not combine token and token type into an enhanced enum (similar to rust's implementation)?

shreyassanthu77:
> If using the new dart language features could we not combine token and token type into an enhanced enum (similar to rust's implementation)? 
Good thought but the problem is dart's enhanced enums aren't like tagged unions ( normal enums in rust ), enhanced enums can have data attached to it's fields ( or whatever you call them IDK ) but they must be constant which means I can't store the name of an identifier or the value of a literal type like string or number that is not known at compile time

D-James-GH:
Ahh yes, of course, I wasn't thinking about dynamic tokens like ident etc...

Fa-C-Shus (approved):
I like separating the Token code and te addition of utils

vhladko:
@shreyassanthu77 is it ready to be merged ? :)

shreyassanthu77:
Yes 👍

Fa-C-Shus:
> Yes 👍

good job; I'm moving over to Functional Dart with pure functions and immutability

vhladko:
Nice, then let’s go :)

shreyassanthu77:
Hey @vhladko can't I merge my own code into the repo ? I don't want to bother you guys every time I have a new pr.

vhladko:
It’s more question to Prime when he is back :)

PR #67, opened 2023-05-28
Add Carbon Implementation

MarvelousAnything:
I finally found a language that was not taken—time to learn Carbon.


MarvelousAnything:
WTF. It seems that there is no Carbon compiler. Only a compiler explorer. I do not feel like writing a Carbon compiler.

PR #28, opened 2023-05-26
Add Swift version

marcusziade:
# Wrote this in Neovim, btw. 
cluelessswift

PR #72, opened 2023-05-28
Added Swift implementation

MatsMoll:
Added some sweet Swift.

Btw, its kind of like traits


marcusziade:
Redundant :)
* [#28]

MatsMoll:
Nice! Let's combine forces on swift 🚀 @marcusziade

PR #76, opened 2023-05-29
Made the PHP version slightly less yucky

unconv:
Made the PHP version a bit cleaner by using backed enums and constructor property promotion. Also made it work with an underscore in the ident.


wartab:
This breaks the parser and makes the language locale dependent.

wartab:
That said I added constructor property promotion in my brand, I didn't know that feature existed.

vhladko:
@wartab @unconv should these changes be merged ? Or some changes needed ?

unconv:
@vhladko I tested my previous changes only with PHP 8.1 and the test passed. I used the actual Dockerfile now and fixed issues in my code.

I think @wartab has developed his interpreter a lot further in his own branch, but hasn't made a pull request yet.

My changes might have just made the code even yuckier so might as well close this one.

wartab:
The backed enum causes issues, because you use the values for parsing and text like INT or FUNCTION shouldn't result in the token they represent.

unconv:
@wartab Yeah, I used those values because PHP's backed enums require all cases to have a backing. The tokens are `tryFrom`ed or `match`ed from single characters, so multi character backed tokens should never result from those. But I guess that's moving from one hack to another...

wartab:
Right, I'm dumb lmao
Implementing the rest I often forgot I'm dealing with just characters.

PR #77, opened 2023-05-29
make python version more readable 

paulfrische:
I've fixed a typo, added python bytecode to the .gitignore and ran a formatter over the code. The functionality/logic hasn't changed.

### DIFF
--- a/python/lexer/lexer.py
+++ b/python/lexer/lexer.py
@@ -39,8 +39,8 @@ class Lexer:
             return tok
 
         if is_letter(self.ch):
-            indent = self.read_ident()
-            return keywords.get(indent, Token(TokenType.Ident, indent))
+            ident = self.read_ident()
+            return keywords.get(ident, Token(TokenType.Ident, ident))
         
         if self.ch.isdigit():
             return Token(TokenType.Int, self.read_int())
--- a/python/lexer/lexer.py
+++ b/python/lexer/lexer.py
@@ -1,13 +1,12 @@
 from tokens import Token, TokenType, char_to_token_type
 
 
-keywords = {
-    'fn' : Token(TokenType.Function, 'fn'),
-    'let': Token(TokenType.Let, 'let')
-}
+keywords = {"fn": Token(TokenType.Function, "fn"), "let": Token(TokenType.Let, "let")}
+
 
 def is_letter(ch):
-    return ch.isalpha() or ch == '_'
+    return ch.isalpha() or ch == "_"
+
 
 class Lexer:
     def __init__(self, input):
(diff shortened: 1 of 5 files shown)


python/lexer/lexer.py
```diff
@@ -1,36 +1,34 @@
 from tokens import Token, TokenType, char_to_token_type
 
 
-keywords = {
-    'fn' : Token(TokenType.Function, 'fn'),
-    'let': Token(TokenType.Let, 'let')
-}
+keywords = {"fn": Token(TokenType.Function, "fn"), "let": Token(TokenType.Let, "let")}
```
thetredev, line 4: this is actually less readable than the removed lines

thetredev: No, PEP8 says your previous version is correct: https://peps.python.org/pep-0008/#indentation. Quote:
> The closing brace/bracket/parenthesis on multiline constructs may either line up under the first non-whitespace character of the last line of list, as in:
```python
my_list = [
    1, 2, 3,
    4, 5, 6,
    ]
result = some_function_that_takes_arguments(
    'a', 'b', 'c',
    'd', 'e', 'f',
    )
```
> or it may be lined up under the first character of the line that starts the multiline construct, as in:
```python
my_list = [
    1, 2, 3,
    4, 5, 6,
]
result = some_function_that_takes_arguments(
    'a', 'b', 'c',
    'd', 'e', 'f',
)
```
Minus the extra whitespace you had before `:`. So PEP8-correct would be this:
```python
keywords = {
    "fn": Token(TokenType.Function, "fn"),
    "let": Token(TokenType.Let, "let")
}
```
That's how I understand it at least.

WoodyWoodster: I personally prefer the multi-line as well. Something to consider is reverting this back to a multi-line and adding a trailing comma to make it compliant with PEP8/black.
``` python
keywords = {
    "fn": Token(TokenType.Function, "fn"),
    "let": Token(TokenType.Let, "let"),
}
```

https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#the-magic-trailing-comma

python/lexer/lexer_test.py
```diff
@@ -6,71 +6,71 @@
...
+            TokenType.Semicolon,
         ]
 
         lex = Lexer(input)
         for token in tokens:
             self.assertEqual(lex.get_next_token().type, token)
 
     def test_get_next_token_2(self):
-        input ='''\
+        input = """\
```
WoodyWoodster, line 27: Could we update this to look like the following:
```python
input = """
    let five = 5;
    let ten = 10;
    let add = fn(x, y) {
        x + y;
    };
    let result = add(five, ten);
"""
```

vhladko:
@paulfrische any updates on this one ? :)

paulfrische:
I'll update this PR. I just don't have time right now.
Have a Great Day!

vhladko:
@paulfrische any updates on this one ? :)

PR #79, opened 2023-05-29
feat(c#) - finish ch1

redducks100:


vhladko:
hey, @redducks100, we appreciate your effort to do that, looks really good ! We already have a csharp folder with implementation, so you could join that team and suggest changes to existed repo, or if you feel that your implementation is different, you could rename the folder to something like csharp2 and add yourself as a CODEOWNER of that repo :)

vhladko:
I’ll close this one as you already have benchmark :)

PR #63, opened 2023-05-27
Add Dlang implementation

ToraKuma42:
Ask and you shall receive!

Only did this for the meme. :)

PR #82, opened 2023-05-29
D lang

lucasnethaj:


ToraKuma42:
Got my implementation here - https://github.com/ThePrimeagen/ts-rust-zig-deez/pull/63
Just letting you know in case you had any questions while writing your implementation.

lucasnethaj:
Ah cool. I tried to look but i couldn't find an existing Implementation. Will close this on now then

PR #85, opened 2023-05-30
cpp: add a Dockerfile

codebrainz:


ThePrimeagen:
i acushually don't know if this is correct, but it does look correct :)

ThePrimeagen (approved):

Issue #84, opened 2023-05-30
Looking for maintainers of specific languages.


ThePrimeagen:
My fellow degens,

As Lord of the degens it's rest upon me to lead us into this glorious interpreted challenge, but at the same time I find myself weary of languages I do not understand. I call upon the degen legion to bring forth representatives willing to shepherd various languages

I will bequeath merge powers to whoever wins the hearts and minds of ethe degens

So arise my maidenless tarnishes, nominate yourself for a language.

All languages that are not TypeScript, rust, zig need representatives.

No, I did not proofread this, why do you ask?

ThePrimeagen:
I will choose at the end of the stream tomorrow based on the most degenerate response per language

May the odds ever be in your favor

alanwilhelm:
I'd be happy to shepherd your coverage of Elixir. A language nobody cares about that does frontend shit nobody even understands. It's actually pretty BA: https://phoenixframework.org/blog/phoenix-liveview-0.19-released

metruzanca:
This thread is going to be glorious. I can't wait.

trvswgnr:
here's the thing: we don't need any other maintainers. as the Head Forker at [CrabLang](https://github.com/crablang/crab) (maybe you've heard of it) i'm uniquely qualified to maintain not just the javascript, typescript, and ~~rust~~ crab implementations, but all other languages as well. all you have to do is hand me the keys to the castle, then sit back and watch the numbers go up.

you won't regret this.

pgaijin66:
I would love to provide support for Go. I am not a pro :P but have built couple of applications. Having said that, i see this as an exciting opportunity to learn from other fellow programmers as well. I will definitely try to give my best :).

Stolkerve:
Hey I'm not qualify, niether speak a good level of english, but I have a good heart <3.

bl42:
Dear Lord of the degens,

As a devoted member of the degen legion, I humbly request permission to maintain the markdown file for our glorious interpreted challenge. I understand the importance of preserving and organizing our collective knowledge, and I would be honored to take on this responsibility.

With your permission, I will ensure the readme.md file remains up to date, accurate, and accessible to all degens. I will diligently manage the content, making necessary edits and additions as needed. Rest assured, I will handle this task with utmost care and attention to detail.

May the degens continue to thrive under your leadership!

Yours faithfully,
Markus the Markdown Keeper

scorbettUM:
[...]

PR #90, opened 2023-05-30
Initial CODEOWNERS file

ryanwinchester:
You might need to update repository settings if you want to do something like add some branch protection rules for codeowners.

For the initial version, I just added the people who contributed the language, except I put myself for Elixir, since that person already conceded it to moi.

Can be updated later with results from #84
You can also have more than one person, separated by space.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,22 @@
+# Global owners.
+@ThePrimeagen
+
+# Degenerate language maintainers.
+/c/             @zivlakmilos
+/cpp/           @nicdgonzalez
+/crystal/       @devnote-dev
+/csharp/        @MarioGK
+/dart/          @shreyassanthu77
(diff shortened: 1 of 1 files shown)


CODEOWNERS
```diff
@@ -0,0 +1,22 @@
...
+@ThePrimeagen
+
+# Degenerate language maintainers.
+/c/             @zivlakmilos
+/cpp/           @nicdgonzalez
+/crystal/       @devnote-dev
+/csharp/        @MarioGK
+/dart/          @shreyassanthu77
+/elixir/        @ryanwinchester
+/erlang/        @mikimaine
```
yordis, line 11:
```suggestion
/elixir/        @ryanwinchester @yordis
/erlang/        @mikimaine @yordis
```

💜

PR #93, opened 2023-05-30
refactor(fsharp) Remove mutations

kram1138:
Makes the code more "functional" F# code as opposed to a direct implementation of the interpreter book

### DIFF
--- a/fsharp/Monkey/Lexer.fs
+++ b/fsharp/Monkey/Lexer.fs
@@ -20,23 +20,30 @@ type Token =
 
 type Code =
     { Source: string
-      mutable Position: int
-      mutable ReadPosition: int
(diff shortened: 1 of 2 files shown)


meenzen:
I like it, especially because this uses a sequence expression instead of my janky list concatenation.

bjornkihlberg:
You could use computation expressions to implement a parser monad.

meenzen (changes requested):
Would you mind rebasing this on main?

vhladko:
@meenzen are you good with changes, can we merge it ?

meenzen (approved):
LGTM

vhladko:
Nice, will merge :)

PR #99, opened 2023-05-31
cpp: rewrite?

Spongman:
c++ code that doesn't suck balls, actually implements the lexer functionality and passes the tests (borrowed from the C version).

_zero_ memory allocations btw.


Sirraide:
Yeah, the current implementation is really just ‘C with classes’ and not C++...

arghasen:
I like this implementation

cpp/src/lexer/lexer.cc
```diff
@@ -0,0 +1,67 @@
...
+		case ')': return { TokenType::rparen    };
+		case ',': return { TokenType::comma     };
+		case '+': return { TokenType::plus      };
+		case '{': return { TokenType::lsquirly  };
+		case '}': return { TokenType::rsquirly  };
+	}
+
+	// parse integer
+	if (isdigit(ch)) {
+		while (position_ < length_ && isdigit(input_[position_]))
```
vgasparyan1995, line 44: This (and the similar lines below) looks like a good place to use string_view::first_not_of.

cpp/src/lexer/lexer.hh
```diff
@@ -0,0 +1,41 @@
...
+	lsquirly,
+	rsquirly,
+	function,
+	let,
+};
+
+
+struct Token final
+{
+	const TokenType type;
```
vgasparyan1995, line 26: This const seems too harsh. It effectively strips away copy-assignability. As a user, I'd expect the Token returned by nextToken() to be a trivially copyable simple struct. Let me take a const Token if I mean it.

cpp/src/lexer/lexer.cc
Spongman: i very much doubt that's more efficient that `is*()`

cpp/src/lexer/lexer.hh
```diff
@@ -0,0 +1,41 @@
...
+class Lexer final
+{
+	public:
+		Lexer(std::string_view);
+		Token nextToken() noexcept;
+
+	private:
+		const std::string_view input_;
+		const size_t length_;
+		size_t position_ = 0;
```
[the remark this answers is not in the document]
Spongman: why update two values (start & length) instead of just one (position) ?

cpp/src/lexer/lexer.cc
Spongman: new version uses a `string_view` and  a `string_view::iterator` only.

vhladko:
@nicdgonzalez can you take a look to this one ? :)

nicdgonzalez (approved):
Looks good to me 👍🏽

vhladko:
Okay, then lets go :)

PR #103, opened 2023-05-31
first chapter complete

boreddad:
arch user here. wanted throw in a version using c89 compiler


ThePrimeagen:
could you add yourself to the codeowners file under your folder?

that way others can commit to this folder and you get required to review :)

PR #107, opened 2023-06-01
refactored Token & Utils Lexer complete through pg 25 -Dart

Fa-C-Shus:
This pull request refactored the Token class to its own file following the repo creator's work. It is also completed through Start a REPL


Fa-C-Shus:
two new commits:
1. includes the REL which really just spits out tokens
2. Parser running through LetStatement and ReturnStatement

PR #96, opened 2023-05-30
Refactored from raw javac to Maven

Crain-32:
Swapped from using `javac` to compile and run, to instead leveraging Maven. Currently fails the Readme Standards in regards to a Makefile/Dockerfile. Hoping someone else makes that change, or I can look into it later.

Decided on the package name `startup.the.monkey`.

PR #108, opened 2023-06-01
Improving the Java solution

RafRunner:
Improving the Java solution by using more modern features (returning from switch, Optional, multi line String, type inference) and native Character methods.

Both tests still pass.

### DIFF
--- a/java/Lexer.java
+++ b/java/Lexer.java
@@ -2,9 +2,11 @@ import java.util.HashMap;
 import java.util.Optional;
 
 class Lexer {
-    String input;
-    int pos;
-    HashMap<String, Token> keywordMap = new HashMap<>() {
+
+    private final String input;
+    private int pos = 0;
+
+    final HashMap<String, Token> keywordMap = new HashMap<>() {
         {
             put("fn", createToken(TokenType.FUNC, "fn"));
             put("let", createToken(TokenType.LET, "let"));
@@ -18,7 +20,7 @@ class Lexer {
     public Token nextToken() {
         this.skipWhitespace();
 
-        TokenType tt = switch (this.getCc()) {
+        final var tokenType = switch (this.getCc()) {
             case '{' -> TokenType.LSQIRLY;
             case '}' -> TokenType.RSQIRLY;
             case '(' -> TokenType.LPAREN;
@@ -31,13 +33,13 @@ class Lexer {
             default -> TokenType.ILLEGAL;
         };
 
-        Token t = switch (tt) {
-            case EOF -> this.createToken(tt, "eof");
-            default -> this.createToken(tt, String.valueOf(this.getCc()));
+        final var token = switch (tokenType) {
+            case EOF -> this.createToken(tokenType, "eof");
+            default -> this.createToken(tokenType, String.valueOf(this.getCc()));
         };
 
         if (Character.isLetter(this.getCc())) {
-            String ident = this.indent();
+            var ident = this.indent();
 
             return Optional
                     .ofNullable(this.keywordMap.get(ident))
@@ -47,7 +49,7 @@ class Lexer {
         }
 
         this.advance();
-        return t;
+        return token;
     }
 
     private void advance() {
--- a/java/Lexer.java
+++ b/java/Lexer.java
@@ -1,15 +1,15 @@
 import java.util.HashMap;
 import java.util.Optional;
 
-class Lexer {
+public class Lexer {
 
     private final String input;
     private int pos = 0;
 
     final HashMap<String, Token> keywordMap = new HashMap<>() {
         {
-            put("fn", createToken(TokenType.FUNC, "fn"));
-            put("let", createToken(TokenType.LET, "let"));
+            put("fn", TokenType.FUNC.createToken("fn"));
+            put("let", TokenType.LET.createToken("let"));
         }
     };
 
@@ -34,8 +34,8 @@ class Lexer {
         };
 
         var token = switch (tokenType) {
-            case EOF -> this.createToken(tokenType, "eof");
-            default -> this.createToken(tokenType, String.valueOf(this.getCc()));
+            case EOF -> tokenType.createToken("eof");
+            default -> tokenType.createToken(String.valueOf(this.getCc()));
         };
 
         if (token.type() != TokenType.ILLEGAL) {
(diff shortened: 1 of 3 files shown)


java/Lexer.java
```diff
@@ -1,9 +1,12 @@
 import java.util.HashMap;
+import java.util.Optional;
 
 class Lexer {
-    String input;
-    int pos;
-    HashMap<String, Token> keywordMap = new HashMap<String, Token>() {
+
+    private final String input;
+    private int pos = 0;
```
xNaCly, line 7: Is there a reason to make all values final?

java/Lexer.java
```diff
@@ -17,55 +20,48 @@ public Lexer(String input) {
...
-        } else {
-            t = this.createToken(tt, String.valueOf(this.getCc()));
-        }
+        final var token = switch (tokenType) {
+            case EOF -> this.createToken(tokenType, "eof");
+            default -> this.createToken(tokenType, String.valueOf(this.getCc()));
+        };
+
+        if (Character.isLetter(this.getCc())) {
+            var ident = this.indent();
```
xNaCly, line 42: I dislike the use of auto or var, as i understand var works like auto and determines the variables type by itself right? [source](https://docs.oracle.com/en/java/javase/20/language/local-variable-type-inference.html) Not really necessary here because its just a string.

java/Lexer.java
```diff
@@ -17,55 +20,48 @@ public Lexer(String input) {
...
-        if (this.isLetter(this.getCc())) {
-            String ident = this.indent();
-            Token keyword = this.keywordMap.getOrDefault(ident, null);
-            if (keyword != null) {
-                return keyword;
-            } else {
-                return createToken(TokenType.IDENT, ident);
-            }
-        } else if (this.isNumber(this.getCc())) {
+            return Optional
```
xNaCly, line 44: Why is this necessary instead of simply returning the found identifier or `ILLEGAL`?

xNaCly (changes requested):

java/Lexer.java
RafRunner: I believe it's a best practice to make all values that shouldn't change final to avoid unnecessary mutation and hacks, like is the default in many other languages. I don't have a strong preference for this, but I got used to putting final on everything since it's my company's default.

java/Lexer.java
RafRunner: The main reason I used var here is to showcase it. Not many people are familiar with modern Java and probably still think about the 1.7 days. It could show its a possibility.

Crain-32:
@RafRunner could you pull down #96 and do these changes? I'm trying to get us into Maven to make the build process easier, and would like the changes. Assuming @xNaCly approves

xNaCly:
> @RafRunner could you pull down #96 and do these changes? I'm trying to get us into Maven to make the build process easier, and would like the changes. Assuming @xNaCly approves

Is a build system really necessary for that, id like to try to keep it simple. What exactly is the benefit of using maven instead of javac?

Crain-32:
More people know how to use Maven vs Javac. Plus bringing in JUnit for testing is a nice QOL improvement.

xNaCly:
> More people know how to use Maven vs Javac. Plus bringing in JUnit for testing is a nice QOL improvement.

More people know maven than the java compiler? I think its overkill, that doesn't really convince me

Crain-32:
> I think its overkill, that doesn't really convince me

I can admit it's maybe more than required, but I think it's better to make writing the code/tests easier. We want contributing to be as easy as possible, proper build systems make that easier.

RafRunner:
> @RafRunner could you pull down #96 and do these changes? I'm trying to get us into Maven to make the build process easier, and would like the changes. Assuming @xNaCly approves

Sure. Adding maven and a testing framework would be great. I'll put it and add to this pull request as soon as I get to my personal computer in a few hours.

RafRunner:
@xNaCly If you don't want to add a build system it's ok, it could be added later and doesn't add much to just a language showcase, but I do believe we should show some modern features to break some misconceptions about java and show how expressive and practical it can be. What do you think about @asafbennatan 's proposal and giving the TokenType enum the responsibility of crating tokens? If you want me to change anything more just tell me.

xNaCly:
> @xNaCly If you don't want to add a build system it's ok, it could be added later and doesn't add much to just a language showcase, but I do believe we should show some modern features to break some misconceptions about java and show how expressive and practical it can be. What do you think about @asafbennatan 's proposal and giving the TokenType enum the responsibility of crating tokens? If you want me to change anything more just tell me.

I think thats a good idea, i don't really what that would look like but im happy to be educated about it. I also think the build system will be necessary later on for getting and using a http server library due to the fact that the benchmarks will be done that way. Im just a litte biased around not using a build system, especially because gradlew is so slow and everything seems to need three xml files for config.

Crain-32:
@xNaCly I can maintain the build system, my PR is actually using a slightly slower build chain since it's Maven, but I know Maven better than I know Gradle, so I went with what was comfortable. I'm thinking we probably merge this, I can update my PR because Merge conflicts are beautiful, and then we move forward. 

From what I know we're fine on the Build Side after that, I'd have to know more about the HTTP expectations, but I do Springboot Development for my job, so I'm not too worried about wrapping it in a Server.

Swapping topics.

Regarding the `TokenType` enum creating Tokens, I'm not the biggest fan. Enums are just constants, the less we give to them, the better.

I think if we swap how it works right now I say we upgrade to Java 20 where we have super cool Pattern Matching that we could show off. It's incredibly powerful, and would fit well into this workflow. We'd just have to `--enable-preview` on the run scripts.
We also have it in Java 17, but that would be Preview 1 instead of Preview 4, an it's even better in Preview 4.
[Preview 1 JEP](https://openjdk.org/jeps/406)
[Preview 4 JEP](https://openjdk.org/jeps/433)

xNaCly:
> I can maintain the build system, my PR is actually using a slightly slower build chain since it's Maven, but I know Maven better than I know Gradle, so I went with what was comfortable.

My Problem is mainly the package things, I hate having to go down three or more levels until i am at the source files. 

> I think if we swap how it works right now I say we upgrade to Java 20 where we have super cool Pattern Matching that we could show off. 

That sounds great, I agree.

xNaCly (approved):

java/Lexer.java
andirady: I'm okay with ``var`` but ``final var`` is unnecessary for me. I personally thinks, you don't need to declare a variable as final in method scope. Since your method is supposed to be small, hence it should be obvious how the variable is being used.

vhladko:
@xNaCly  @Crain-32 @RafRunner  is this PR ready to be merged ? I've seen that it holds another PR #96

java/Lexer.java
RafRunner: I removed final from local variables.

java/Lexer.java
RafRunner: @andirady I removed final from local variables. It does cause some unnecessary noise.

RafRunner:
> @xNaCly @Crain-32 @RafRunner is this PR ready to be merged ? I've seen that it holds another PR #96

I believe so, I've addressed all points raised.

vhladko:
Okay, then merged :)

PR #113, opened 2023-06-01
R5RS (())(()()()))(((

Fesaa:
Used this dialect of Scheme once in University, thought it would be funny.
It's far from done yet, and misses a few things...

The final plan would be to make it as vanilla scheme as possible, so still have to implement hash map without rackets help.  
Still getting back into the language. So; will probably rewrite a bunch later on to make it more readable / concise

Also, why does my .gitignore never ignore .DS_Store 😭 

Anyway, yeah, scheme ig


ThePrimeagen:
.DS_Store remove pls, add yourself to code owners

Fesaa:
This is all done apart from chapter four and the custom hash maps. Docker runs just runs the test scripts. Dunno what it should be displaying but just does this now 🤷‍♀️ 

image

PR #115, opened 2023-06-01
csharp: Changed to use more language features

TomasDurica:
I'm late to the party as usual.
tl;dr; Tom is a genius. 

I am creating this PR for folks, who want to see csharp in a bit of a better light.
I applaud @MarioGK for delivering code that gets the job done, and is not that bad.
I've noticed that the solution didn't use a lot of language features, that make solving the problem really easy.  Btw. using IEnumerable was a great choice

1. I've represented the tokens as a poor mans union / enum struct (similar to Rust) until the language will have a native support. This solution is exceptionally great in conjunction with advanced pattern matching, which will help in later sections of the book
2. I've used `ReadOnlySpan<char>` to pass around the currently parsed substring - helping the performance without allocations.
3. Advanced pattern matching for multichar tokens and identifiers
4. Snapshot testing - which I believe is amazing match for this kind of project.

Feel free to ask questions directly in PR - I will be happy to help / adjust the implementation.


leo-costa:
Nice @TomasDurica. I also put out a PR but I kept the implementation as close as possible to the Rust one. My initial thought was to use `ReadOnlySpan<char>` but that would work on inside `async` methods.  My implementation was about 30% faster the current one and 60% better memory usage because of some usage of `ReadOnlySpan<char>`.

TomasDurica:
@leo-costa - I don't know if I was just dumb or sleep deprived to search only for "csharp" and not "c#" - so I haven't noticed. Ooopsies :D.

I wanted to go more for idiomatic solution - at least in some places as the `record` trick is probably not that widespread, as seeing just "clones" of C everywhere is not getting the point of this across I feel.

I like your solution, if I'd saw it, I'd maybe not do mine (or at least not with a PR), especially with the `Memory`, to combat the limitations of `ReadonlySpan`. 

I'd probably merge the ifs into the switch in `NextToken`, but that's just my weird personal preference, as I am usually the only one in the room that finds it more readable :D.

Regarding benchmarks:
 |             Method |     Mean |     Error |    StdDev |   Gen0 |   Gen1 | Allocated |
 |------------------- |---------:|----------:|----------:|-------:|-------:|----------:|
 | BenchLexerOldSmall | 1.241 us | 0.0070 us | 0.0062 us | 0.6294 | 0.0019 |    2.9 KB |
 | BenchLexerNewSmall | 1.859 us | 0.0276 us | 0.0258 us | 0.3681 |      - |    1.7 KB |
 |   BenchLexerOldBig | 2.511 us | 0.0149 us | 0.0132 us | 1.2550 | 0.0076 |   5.78 KB |
 |   BenchLexerNewBig | 3.752 us | 0.0111 us | 0.0098 us | 0.5951 | 0.0038 |   2.74 KB |
I am a bit slower - I'd guess the pattern matching on span with it being slowly rolled is not the fastest thing, but I feel the readability is worth the additional micro-seconds (even though mine is 1,5 the speed of the original). The fact that my solution uses less memory is probably not that surprising, as I am not storing the string representation of the token if it is not needed.

Anyway - very nice solution indeed, and apologies for duplication then.
Just wondering - why `struct` for the `Lexer`? Is it only to match Rust, or is there another benefit?

vhladko:
As I see it's a completely new version of implementation that potentially could  have a difference in a perf and readability, if you want, you can create a csharp2 folder (as it is for kotlin2) and add yourself as a CODEOWNER of that version :)

leo-costa:
>Regarding benchmarks:

What input did you use to test yours? I would like to run it against mine and check it out too!

> I'd probably merge the ifs into the switch in `NextToken`, but that's just my weird personal preference, as I am usually the only one in the room that finds it more readable :D.

I actually had it like that initially, but as you know C# doesn't allow for (without ugly code) multi line expression on a switch arm 😢 . I had a solution just like your, using tuple where the second field is the amount of `ReadChar`. I wanted to keep as close as possible to the Rust one. I guess I'll revisit that again.

> Anyway - very nice solution indeed, and apologies for duplication then.

No need to apologize you got here first 👍 .

> Just wondering - why struct for the Lexer? Is it only to match Rust, or is there another benefit?

Do you mean `struct` for the `Token`? if that's the question, then it's because there would be no memory allocation when returning that struct from `NextToken()`

TomasDurica:
@leo-costa: And now a double oopsie :D. I was reading the [PR](https://github.com/ThePrimeagen/ts-rust-zig-deez/pull/79) of @redducks100 which is actually quite similar to yours.
The question regarding struct was that in the other solution, the Lexer itself was a struct, what spiked my curiosity - any ideas why that may be beneficial? Having Tokens as structs make sense - I will need to fiddle a bit with record structs - will see if that helps
Regarding benchmark - I've used the two test inputs that were present - feel free to benchmark mine against yours in any input - I will be interested in the results (I am pretty sure I will loose in all cases :D )

@vhladko I believe this to be a bit unique approach, as I am heavily relying on the pattern matching (currently working on chapter 2 and oh boy, you can definitely feel the difference) and "stateless" parsing. I'd like to present this as a separate approach, and maybe @leo-costa may chime in? I definitely plan to complete the whole solution and stick to the concept. What do you both think? Does it deserve a separate directory (csharp2 or csharp-patterns)?

vhladko:
You can name new folder as you want, but if you believe it deserve separate repo go ahead, don’t forget to add yourself as Codeowner :)

vhladko:
@TomasDurica we’ve just merged PR by @leo-costa, what do you think, do you want to help his implementation or have it as csharp3 ? It’s only your decision :)

redducks100:
> @leo-costa: And now a double oopsie :D. I was reading the [PR](https://github.com/ThePrimeagen/ts-rust-zig-deez/pull/79) of @redducks100 which is actually quite similar to yours. The question regarding struct was that in the other solution, the Lexer itself was a struct, what spiked my curiosity - any ideas why that may be beneficial? Having Tokens as structs make sense - I will need to fiddle a bit with record structs - will see if that helps Regarding benchmark - I've used the two test inputs that were present - feel free to benchmark mine against yours in any input - I will be interested in the results (I am pretty sure I will loose in all cases :D )
> 
> @vhladko I believe this to be a bit unique approach, as I am heavily relying on the pattern matching (currently working on chapter 2 and oh boy, you can definitely feel the difference) and "stateless" parsing. I'd like to present this as a separate approach, and maybe @leo-costa may chime in? I definitely plan to complete the whole solution and stick to the concept. What do you both think? Does it deserve a separate directory (csharp2 or csharp-patterns)?

I would assume there wouldn't be much difference between declaring the Lexer as a struct or class as there shouldn't be or we don't expect more than 1 instance anyway.  When I wrote my implementation, it was purely left over when I was experimenting with ReadOnlySpans variables. Also your advanced pattern matching looks really cool, learned something new. I would love to benchmark struct/readonly struct/record/record structs, but need to find the time for that.

MarioGK:
Hi @TomasDurica the genius,
Your implementation seems to be more readable and faster than mine just talk with @leo-costa and maybe merge your implementation together?
But feel free to take over the csharp implementation, i am not having much free time to contribute.

vhladko:
@MarioGK @TomasDurica so we can merge current PR as changes to csharp repo and add Tom as a second owner for this folder, what do you think ?

TomasDurica:
@redducks100 : Ah now it make sense - I guess you also got struct by the Span limitations (especially in the iterator blocks). I was actually not aware you can't do that :D. Happy to be of help, Feel free to contribute / ask if you spot something :) I hope I will have chapter 2 done today - I think I took it too far with that, but will see - anyway, I will also try to see if I can improve performance, and will include a bench project like you did.

@vhladko: If @leo-costa and @redducks100 will not have any problems with it (I am happy to swap if they want to keep #1)

leo-costa:
> @redducks100 : Ah now it make sense - I guess you also got struct by the Span limitations (especially in the iterator blocks). I was actually not aware you can't do that :D. Happy to be of help, Feel free to contribute / ask if you spot something :) I hope I will have chapter 2 done today - I think I took it too far with that, but will see - anyway, I will also try to see if I can improve performance, and will include a bench project like you did.
> 
> 
> 
> @vhladko: If @leo-costa and @redducks100 will not have any problems with it (I am happy to swap if they want to keep #1)

You can be #1 🚀👊

vhladko:
@leo-costa has his own repo and from what I see, everyone pretty much like your solution, so I’ll merge this one, could you add yourself as a owner for that folder ? @redducks100  I gonna ask you to have separate folder and make yourself a Codeowner of that repo csharp3 as well, cause as I understood your version could show different result from this one :)

vhladko:
@TomasDurica pls add yourself as a Codeowner of that folder :)

TomasDurica:
Will do :) thanks

vhladko:
Let’s go :)

PR #118, opened 2023-06-01
Rust: Changing Strings to &str and Vec<u8> to &[u8]

joseburgosguntin:
Also, todo!() to Token::Illegal, tidding up the test, and adding an Token::Illegal to the test


ThePrimeagen:
so i don't know if i want to change anything yet because i would love to do a perf review and change these things (such as string copies vs references, etc etc)

ThePrimeagen:
i would literally keep the changes with the use Token::* (didn't know that) and illegal, just no perf based ones yet :)

PR #119, opened 2023-06-01
Scala 3

krined-dev:
An implementation in Scala 3 using new language features such as the new Enums for ADTs and extension methods using the extension keyword. I tried to keep the solution as close as possible to the book while using Scala language features.


novusnota:

vhladko:
Can you add yourself as a Codeowner ? And check if your code can be easily run without docker or makefile :)

vhladko:
also, we've just merged one Scala implementation, feel free to help existed repo to be better or rename a folder to scala2 if you want to keep your implementation :)

PR #127, opened 2023-06-02
Refactor odin lexer

fishcken:
First of many changes I want to make to the Odin lexer. Here we can spare the `fn` and `let` procedures as we can just read them like a normal identifier. We also use a `skip_whitespace`  so that we can skip all the whitespace in a single call to `scan_token`.


fishcken:
Damn, I just noticed the code was updated, I forked this just yesterday

PR #128, opened 2023-06-02
Refactor Odin lexer

fishcken:
Small improvement to the Odin code.

I believe this better showcases Odin's capabilities by making better use of the switch-case construct. I also removed a few procedures (`number`, `keyword`, `ident`), as I feel the code is rather simple and thus we don't really need to abstract it away (we can add them back if complexity increases later), we also improve readability by keeping all `append(&l.tokens, ...)` calls to a single place (the `scan_token` procedure).

Hopefully you can agree with me :)

### DIFF
--- a/odin/lexer/lexer.odin
+++ b/odin/lexer/lexer.odin
@@ -90,7 +90,6 @@ scan_token :: proc(l: ^Lexer) {
 	start := l.offset - 1
 	//odinfmt: disable
 	switch c {
-	// Single Char:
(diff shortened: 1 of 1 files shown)


vhladko:
@jon-lipstate can you take a look to this one ? :)

fishcken:
Just noticed we're mixing tabs and spaces in my commit, definitely we should fix that if we merge, no clue how github works for that tho

PR #131, opened 2023-06-02
Scala: Cleanup .gitignore

alextsc:


vhladko:
Sorry, I've missed that gitignore is something from template, so let's keep it as it is for now, my bad :(

PR #133, opened 2023-06-02
[Nim] Add basic unit testing

Echriser:
This test suite currently only checks the enums, not the finalized strings [Ex: "Ident ten"].


vhladko:
Looks good, will merge it :)

PR #135, opened 2023-06-02
Add support for HolyC.

xslendix:


vhladko:
Looks like there is a conflict in a Codeowner file, can you resolve it so I can merge ? :)

xslendix:
Before you merge, can you please test Dockerfile? I cannot get docker to run so I did it blindly.

vhladko:
Then we have to wait till tomorrow :) my tomorrow, I believe we have different time zone :)

xslendix:
Yeah that's fine lol

xslendix:
I was able to get Docker to run and I have fixed everything there was to fix. I also left a README file explaining more in-depth how all this works. I also created a fork of 3Days to work without audio or SDL2. This is not only faster to build, but also usable in Docker.

vhladko:
Cool ! Then I think we can merge it now :)

PR #142, opened 2023-06-03
Nim changes

Nmerryman:
This PR has two small changes.
First I removed the testing code from `lexer.nim` because it already exists in the `tests` directory.
Second I made minor but easy changes to the text to bring it closer to the public style guides.

### DIFF
--- a/nim/.gitignore
+++ b/nim/.gitignore
@@ -1,0 +1,1 @@
+tests/test.exe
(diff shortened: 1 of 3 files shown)


Echriser:
Looks good, I approve.

Nmerryman:
Just wondering, why didn't you make full `Token(tokenkind, litteral)` and a `Lexer` type?

Echriser:
> Just wondering, why didn't you make full `Token(tokenkind, litteral)` and a `Lexer` type?

Simply took the most straightforward path I could think of since I wrote the entire thing during the lexer review stream. I might make/approve refactors later if there's enough benefit, but for now my book is still being shipped.

PR #146, opened 2023-06-03
More changes to the lexer

Nmerryman:
Cleaned up the code a bit and made it nicer to use/look at.

All of my changes were made while referencing 3 alternative nim implementations (jakelsen, lantos1618, and mine).

We probably need to overhaul the lexer to be much closer to what the book does because there are details that were carefully planed for future use that got discarded. Tests still work of course.

### DIFF
--- a/nim/src/lexer.nim
+++ b/nim/src/lexer.nim
@@ -2,19 +2,19 @@ import std/[strutils, sequtils]
 
 type 
   TokenKind* = enum
-    ident = "Ident"
-    t_int = "Int"
-    illegal = "Illegal"
-    equal = "Equals"
-    plus = "Plus"
-    comma = "Comma"
-    semicolon = "Semicolon"
-    leftparen = "LeftParen"
-    rightparen = "RightParen"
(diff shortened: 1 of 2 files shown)


Nmerryman:
I noticed you said your book hadn't arrived yet, should I be less aggressive with my PRs?

vhladko:
@Echriser could you look on this one ? Btw did you get automatic notification about this PR cause changes are to your owner folder ?

Echriser:
> I noticed you said your book hadn't arrived yet, should I be less aggressive with my PRs?

It's fine. I appreciate the cleanup work. The changes here all look good to me.



> @Echriser could you look on this one ? Btw did you get automatic notification about this PR cause changes are to your owner folder ?

This mention is the only thing that appeared in my email inbox and not the initial PR, so I assume not.

vhladko:
Hm, that’s strange, I would assume that as a Codeowner you should get notification when there are changes to your folder, will work on it :)

vhladko:
@Echriser as it looks good to you, I’ll merge PR :)

PR #147, opened 2023-06-03
Nim changes

Nmerryman:
This should transition the shape of the lexer to be much more similar to how the book has it. Once again I used inspiration of all three other repos where they were useful to try to build it with an average syntax. At this point the lexer is just about caught up with the PRs others have made and I think this puts us about half way through ch1. Now I sleep.

@Echriser  :)

### DIFF
--- a/nim/src/lexer.nim
+++ b/nim/src/lexer.nim
@@ -4,19 +4,12 @@ import std/[strutils, sequtils]
 type 
   TokenKind* = enum
     tkIllegal
-<<<<<<< HEAD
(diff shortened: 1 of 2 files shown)


Nmerryman:
@Echriser I am pinging once again, pleading with you

Echriser:
Sorry for the wait, looks good. My book should arrive tomorrow so I should be more useful by then.

vhladko:
Okay, then I’ll merge this one :)

PR #152, opened 2023-06-03
Common Lisp: Finishing up the unfinished

lh2:
This finishes up all the unfinished stuff from the first PR.
Beyond a couple of small bugfixes here and there the most notable things are:

- {} blocks now use the same function for statement parsing as the toplevel, which fixes let statements in functions and if's
- Added all the tests for the extentions from chapter 6
- Added support for hash literals
- Added a CLI interface
- Added README


vhladko:
Nice, let’s go :)

PR #153, opened 2023-06-03
using pattern matching instead of map , using singleton Tokens for li…

asafbennatan:
using pattern matching instead of map , using singleton Tokens for literals


vhladko:
@Crain-32 @xNaCly when you have a minute, could you review ? :)

Crain-32:
@vhladko It looks good to me. I'm missing the write access to the Repository from @ThePrimeagen, or I would approve the workflow.

vhladko:
Okay, then I merge :)

PR #155, opened 2023-06-03
feat: create brainf*ck lexer

sno2:
It took longer than expected for me to finish this because some of the memory was unexpected in some places because of scratch cells which ended up causing issues in a few tests.

I guess we might be able to pipe this to a `parser.b` file which can read the tokens out using the binary format later. However, that will be a lot more work than this.


vhladko:
Pls add yourself as a Codeowner :)

sno2:
The codeowners don't work unless if we have write perms.

vhladko:
I know, but it will be changed in future, right now we just need to keep track of codeowners, so pls, add yourself :)

sno2:
ok, I pushed the CODEOWNERS change

vhladko:
Then we can go :)

PR #156, opened 2023-06-03
Chez Scheme implementation of lexer

bjornkihlberg:
Chez Scheme is supposed to be one of the fastest dynamically typed languages so it would be very interesting to see it compared with other solutions with maximum optimization setting.

- Run without compiling
  ```
  $ scheme --script lexer.scm
  ```
- Compile with optimizations
  ```
  $ scheme --script compile.scm --optimize-level 3
  ```
- Run compiled program
  ```
  $ scheme --program lexer.so
  ```
- Pipe source to lexer
  ```
  $ echo let x = 5 | scheme --program lexer.so
  Let
  Ident x
  Equal
  Int 5
  Eof
  ```

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -8,6 +8,7 @@
 /c/             @zivlakmilos
 /c-boreddad/    @boreddad
 /chatgpt/       @unconv
+/chezscheme/    @bjornkihlberg
 /cobol/         @kkard2
 /common-lisp/   @lh2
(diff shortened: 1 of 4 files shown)


vhladko:
Pls add yourself to Codeowner file :)

bjornkihlberg:
> Pls add yourself to Codeowner file :)

Thanks for reminding me!

vhladko:
Nice, let’s go :)

PR #157, opened 2023-06-03
Add Dockerfile, Makefile, implement keywords let and fn

attilakapostyak:
What's new:
- Restructure code (src directory)
- add unit tests
- Add Dockerfile
- Add Makefile
- Implement keywords: let and fn


vhladko:
Nice, will merge it :)

PR #159, opened 2023-06-03
Fix CODEOWNERS path

ThenWhenceComethEvil:


vhladko:
Sorry, that was my bad, thanks for fix :)

PR #162, opened 2023-06-03
[x64_nasm/Refactor] Vastly unscuff file I/O

Sirraide:
Squashed commit of the following:
    [Refactor] Stat file before reading it
    [Refactor] Improve REPL and file reading
    [Refactor] General cleanup


Sirraide:
I saw some discussion about this: I’m in the CODEOWNERS file for this directory, so am I supposed to be able to merge this pr myself since it only affects files in that directory? Because I don’t seem to be able to do so

vhladko:
Yes, in future you will be able, but for now I’ll merge it :)

PR #164, opened 2023-06-03
feat(Java) Implementing section 1.4 of the book

RafRunner:
Implementing section 1.4 - Extending our Token Set and Lexer, following the patterns already implemented.

This sections add a few more keywords and one char tokens, which are trivial to implement, and two two char Tokens (EQUALS and NOT_EQUALS) which are also easy but not so trivial.


java/monkey/src/main/java/Lexer.java
```diff
@@ -20,33 +18,58 @@ public Token nextToken() {
             case ',' -> TokenType.COMMA.token();
             case ';' -> TokenType.SEMI.token();
             case '+' -> TokenType.PLUS.token();
-            case '=' -> TokenType.EQUAL.token();
+            case '=' -> {
```
Crain-32, line 21: I'm thinking we swap this to
```java
case '=' when peekNext() == '=' ->  {this.advance(); yield TokenType.EQUAL.token();}
case '=' -> TokenType.ASSIGN.token();
case '!' when peekNext() == '=' -> {this.advance(); yield TokenType.NOT_EQUAL.token();}
case '!' -> TokenType.BANG.token();
```

vhladko:
@xNaCly @Crain-32 could you take a look ? :)

java/monkey/src/main/java/Lexer.java
```diff
@@ -59,10 +82,14 @@ private void advance() {
     }
 
     private char getCc() {
-        if (this.pos >= this.input.length() || this.pos < 0) {
+        return this.peek(this.pos);
+    }
+
+    private char peek(int pos) {
```
Crain-32, line 88: Can we add in this function?
```java
private char peekNext() {
    return peek(this.pos + 1);
}
```

Crain-32 (changes requested):

java/monkey/src/main/java/Lexer.java
RafRunner: I sadly don't think

```java
case '=' when peekNext() == '=' ->  {this.advance(); yield TokenType.EQUAL.token();}
``` 

is valid syntax. Patter matching can only be done when capturing the value like:

```java
case Character c when c == '=' && peekNext() == '=' ->  {this.advance(); yield TokenType.EQUAL.token();}
```

and for that to work all branches would have to begin with `case Character c where c == '...'` and I believe that would be too verbose and less clear then an if in the special cases

java/monkey/src/main/java/Lexer.java
RafRunner: Of course! It makes things clearer. Added in my last commit.

Crain-32 (approved):

RafRunner:
I took the time and condensed all checks into a single switch statement to make it more direct. @Crain-32 @xNaCly @vhladko

java/monkey/src/main/java/Lexer.java
```diff
@@ -85,31 +73,33 @@ private char getCc() {
...
-        return peek(this.pos + 1);
-    }
-
     private boolean isLetter(char c) {
         return Character.isLetter(c) || c == '_';
     }
 
     private String number() {
-        int pos = this.pos;
+        int pos = this.pos - 1;
```
Crain-32, line 94: Not the biggest fan of this change. I'd rather look forward than backwards. In my mind the Lexer is an `Iterator<Token>`, and so any data behind us is gone, and we can only glance at the next one.

java/monkey/src/main/java/Lexer.java
```diff
@@ -40,18 +42,10 @@ public Token nextToken() {
             case '>' -> TokenType.GT.token();
             case '<' -> TokenType.LT.token();
             case '\0' -> TokenType.EOF.token();
-            default -> TokenType.ILLEGAL.createToken(String.valueOf(this.getCc()));
-        };
 
-        if (token.type() != TokenType.ILLEGAL) {
```
Crain-32, line 29: Could we accomplish this without the `pos -1`?

java/monkey/src/main/java/Lexer.java
RafRunner: I think this makes more sense than what we were doing before, that is, selectively advancing the pos pointer depending on what token was being generated. That was needed because the identifier and number generation moved the pointer but the others didn't. Now we're following the book more closely: we consume the current char (read and move pos) at the beginning of nextToken, so we don't have to move it later. The `pos - 1` is only needed because the position pointer has already been moved once before starting to create the identifier or number string. Notice we are not moving the pointer (not going backwards) just creating a variable that represents where the string should being.

java/monkey/src/main/java/Lexer.java
RafRunner: We could write
```java
private String indent() {
        int pos = this.pos;
        while (this.isLetter(this.getCc())) {
            this.advance();
        }
        return this.input.substring(pos - 1, this.pos);
    }
```

Or go back to what we were doing before (selectively advancing the pointer) but my preference for this new way in the comment above.

RafRunner:
Before solving the conflicts, should I revert the last commit? Or implement the number and ident functions taking the first char as to not use pos -1?

Btw, is the solution being coordinated in a discord server or anywhere else external? I would hate to be the one going against the current lol

Crain-32:
@RafRunner I wanted to avoid making any executive decisions, but yeah, please revert back to `pos + 1`. I'm a bigger fan of it personally, and it's closer to the book. I'm sorry.

Regarding the coordination, currently we have no system. I wouldn't mind creating a quick readme with all the chapters so people can "reserve" parts of the book.

RafRunner:
@Crain-32 alright, I can't do it now, but I'll be able to commit in a few hours. Full revert or change the functions to receive the first char and remove the - 1? In the second case, we'll be able to keep everything in a single switch and don't advance selectively

Crain-32:
@RafRunner just removing the `-1` logic. Should be good, I'd have to see the result, and I'm fine waiting for it.

RafRunner:
@Crain-32 fixed the merge conflicts and changed the implementation not using pos - 1 👍🏽

Crain-32 (approved):

PR #171, opened 2023-06-04
Add lexer for deez jdsl

Sergushka:


Crain-32:
How is this different from the current 2 Kotlin Repos?
image

vhladko:
@Sergushka do you want to keep it as separate version or it’s almost the same as two previous ? I’m not that kotlin developer so can’t judge :)

Sergushka:
I'll continue working on the parser so it will be clearly a different approach even if it's the same language

vhladko:
Okay, then pls add yourself as a Codeowner :)

Sergushka:
I see there is only one kotlin-multiplatform with 3 owners in CODEOWNERS

vhladko:
If you go with your own folder, you need to add your folder name and yourself as a Codeowner of that repo

PR #176, opened 2023-06-04
Added Docker image for Swift

MatsMoll:
Setup a docker image solution for Swift, and with a basic CLI at the same time


vhladko:
@marcusziade could you take a look to this ? :)

vhladko:
Then, I guess we can also merge this one :)

PR #178, opened 2023-06-04
kotlin-multiplatform: update trailing comma rule in editorconfig

hermannm:
Linting currently fails in `kotlin-multiplatform`, due to a trailing comma rule in our `.editorconfig`. It wants to reformat this:
```
Div({
    classes(ReplStyleSheet.replLayout)
}) {
```
...to this:
```
Div({
    classes(ReplStyleSheet.replLayout)
},) {
```
...which is just horrendous. So this PR disables that rule, and reformats `LexerTest` to comply.

### DIFF
--- a/kotlin-multiplatform/.editorconfig
+++ b/kotlin-multiplatform/.editorconfig
@@ -2,4 +2,4 @@
 
 [*.{kt,kts}]
 max_line_length = 100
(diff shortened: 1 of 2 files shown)


hermannm:
Found a better solution, putting up a new PR

PR #179, opened 2023-06-04
kotlin-multiplatform: switch ktlint gradle plugin to kotlinter

hermannm:
Linting currently fails in `kotlin-multiplatform`, due to a trailing comma rule in our `.editorconfig`. Our linter wants to reformat this:
```
Div({
    classes(ReplStyleSheet.replLayout)
}) {
```
...to this:
```
Div({
    classes(ReplStyleSheet.replLayout)
},) {
```
...which is just horrendous. This is because our current plugin `org.jlleitschuh.gradle.ktlint` uses an old version of `ktlint`, which does not work well with trailing commas. This PR switches the linter to `org.jmailen.kotlinter`, which uses a new, better version of `ktlint`.

### DIFF
--- a/kotlin-multiplatform/.editorconfig
+++ b/kotlin-multiplatform/.editorconfig
@@ -2,4 +2,3 @@
 
 [*.{kt,kts}]
 max_line_length = 100
-ij_kotlin_allow_trailing_comma_on_call_site = true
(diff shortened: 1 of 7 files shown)


ekgame:
Can confirm that linting now passes without the weird issues. LGTM!

vhladko:
Nice, let’s go :)

PR #183, opened 2023-06-04
Refactor haskell lens implementation using State operators

ambroslins:
There are specific lens operators for the state monad. I think this implementation is a bit clearer and better shows the power of lenses.
However I would like to know what @alexjercan, @newmaidumosa and @Ofenhed think about this before I update the README.

### DIFF
--- a/haskell/README.md
+++ b/haskell/README.md
@@ -1010,29 +1010,47 @@ newLexer i = Lexer
     }
 ```
 
-The `peek`, `current` and `advance` functions will look like the following
+Since our `LexerT` is an instance of `MonadState` we can use the combinators
+and operators provided by the `lens` library.
+One of these operators is `use :: MonadState s m => Lens s a -> m a` which
+simply gets the current state and returns the focused field.
+Therefore the `peek` and `current` functions will look like the following
 
 ```haskell
 peek :: LexerT Char
 peek = do
-    lexer <- get
-    pure $ fromMaybe '\0' ((lexer ^. input) BS.!? (lexer ^. readPosition))
+    readPos <- use readPosition
+    buffer <- use input
+    pure $ fromMaybe '\0' (buffer BS.!? readPos)
 
 current :: LexerT Char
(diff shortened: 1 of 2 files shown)


vhladko:
@alexjercan could you take a look to this ? :)

alexjercan:
I added some comments with some fixes. Also the rest of the code looks good. Pretty interesting operators :)

alexjercan:
can you also do `make fmt` just to use the same formatter everywhere?

Ofenhed:
Not having used lenses much (as I think they produce confusing code) I can only agree with @alexjercan. The diff was too big (just looking on my phone it looked as if a lot was pure styling), but the operators were very interesting 😊

ambroslins:
Sorry i forgot to run `make fmt` on the last commit. Should be fixed now.

> I added some comments with some fixes. Also the rest of the code looks good. Pretty interesting operators :)

Where did you add these comments? I can't find them.

I should be able to update the README within the next 24h.

Ofenhed:
@alexjercan Did you remember to commit the comments? They are all stuck as pending until you send them as either plain comments, accept review, or reject review.

alexjercan:
> Where did you add these comments? I can't find them.

Oh you are right, they were pending. No worries, it was related to the imports, it is all good now.

vhladko:
Is it good for merge now ?

Ofenhed:
The README references the old operators, so I would say it should be updated before it's merged. A simple explanation of how `.=` relates to the already explained `a & b .~ c` would probably be enough.

alexjercan:
> Is it good for merge now ?

will ping you when it is ready

ambroslins:
@alexjercan, @Ofenhed. I have updated the README. Feel free to leave suggestions.

And thank you for the feedback on the implementation.

alexjercan:
Yeah, it looks good to me. @vhladko I think it is ok to merge it

PR #97, opened 2023-05-31
chore(php): Add composer and phpunit

arbezerra:
Refactored the code to use composer autoload and phpunit for tests.

PR #185, opened 2023-06-05
mlocati for php extensions

exilesprx:
Extension installer for things like composer, xdebug, etc while keeping docker image php centric.

Assumes composer becomes a dep via #97


vhladko:
@wartab could you take a look ? :)

PR #196, opened 2023-06-05
Update pr-worthy.yml

novusnota:
Pinning down precise version of tj-actions/changed-files action


novusnota:
Fixed issues with pr-worthy workflow, now works as expected

Spongman:
regression in `changed-files` ? great!

PR #201, opened 2023-06-06
Visual Cobol lexer base

LTR-Travis:
This is the start of a Micro Focus Visual COBOL for .NET 6.0 implementation of the compiler project. This handles everything in sections up to and including 1.3 (I'm starting on 1.4 today).

Unfortunately, GnuCobol only fully implements up to the COBOL 85 standard and I'm more comfortable using the feature set from the 2002 standard, so I went with a different variant. Granted, the compiler is available for free, but I doubt anyone else would want to go through the hassle, so I'm including my build folder.

An important note is that this *does* require the .Net 6.0 runtime to be installed (or newer). I make use of standard functions in the System.Char and System.String classes and also use the System.Collections.ArrayList class in my tests.

I will likely be a bit slower to implement, given this is my first COBOL project of anything more than a single procedure.


LTR-Travis:
I think I've added it? Forgive my ignorance if that didn't do it. This is my first time contributing to a public repo.

PR #202, opened 2023-06-06
kotlin-multiplatform: change package name

hermannm:
Changing the package name of the `kotlin-multiplatform` implementation, since I'm working with @ekgame and @LizAinslie now, so it doesn't make sense to use my domain anymore.


ekgame:
The tests are passing, LGTM!

PR #204, opened 2023-06-06
add k lexer impl

fluentpwn:
pretty sure this docker file won't run lmao
will add a state machine impl some time later, maybe


vhladko:
Pls add yourself to Codeowner

fluentpwn:
@vhladko done :)

PR #211, opened 2023-06-08
feat(java): Implementing up to chapter 2.5: The basic parser structure

RafRunner:
This is just a rough implementation where I tried to stay pretty close to the book. I don't really like how I did many things, especially overwriting the toString() method to represent Nodes as strings and abstract classes and multiple inheritance, but I just wanted to throw a quick implementation and take some suggestions.

We can't forget the real complicated part is still ahead so we shouldn't implement some super fancy thing that will make things a lot harder later.


vhladko:
@xNaCly @Crain-32 what do you guys think ? :)

java/monkey/src/main/java/root/Parser.java
```diff
@@ -0,0 +1,104 @@
...
+
+    private Optional<Statement> parseStatement() {
+        return switch (this.currentToken.type()) {
+            case LET -> this.parseLetStatement();
+            case RETURN -> this.parseReturnStatement();
+            default -> Optional.empty();
+        };
+    }
+
+    private Optional<Statement> parseReturnStatement() {
```
Crain-32, line 54: I believe you're over using the `Optional` Class here.

We're always in control of the context, and we know that we'll either be dealing with a Stream of Tokens, or an end of Stream. I don't believe we need to wrap our logic in such a way.

java/monkey/src/main/java/root/Parser.java
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        return Optional.of(returnStatement);
+    }
+
+    private Optional<Statement> parseLetStatement() {
+        LetStatement letStatement = new LetStatement(this.currentToken);
+
+        if (!this.expectPeek(TokenType.IDENT)) {
+            return Optional.empty();
```
Crain-32, line 71: Multiple return statements in a function is a code smell. Can we flatten this? Pattern Matching might help.

java/monkey/src/main/java/root/REPL.java
```diff
@@ -0,0 +1,31 @@
...
+import java.util.Scanner;
+
+public class REPL {
+
+    private static final String PROMPT = ">> ";
+
+    public void start(InputStream inputStream, PrintStream printStream) {
+        var scanner = new Scanner(inputStream);
+
+        while (true) {
```
Crain-32, line 14: I'd swap some stuff around here.

The biggest issue is that you're declaring `while (true)`, but also have a break inside of the loop. This means _we do not loop forever_.

I believe a much better result would be a `do-while` loop, where `line.trim().length() == 0` sets a boolean to `false`, to end the loop.

java/monkey/src/main/java/root/ast/Identifier.java
```diff
@@ -0,0 +1,18 @@
+package root.ast;
+
+import root.Token;
+
+public class Identifier extends Expression {
```
Crain-32, line 5: This is incorrect to the Book.
We can see on Page 45 that Identifier != Expression, as they branch on the `LetStatement` in the AST.

It should _not_ extend Expression.

java/monkey/src/main/java/root/ast/Statement.java
```diff
@@ -0,0 +1,7 @@
+package root.ast;
+
+public abstract class Statement extends Node {
+    public void statementNode() {
```
Crain-32, line 4: Prefer this to be `public abstract void statementNode();` so that implementations are forced to define it, instead of an empty call.

java/monkey/src/main/java/root/ast/LetStatement.java
```diff
@@ -0,0 +1,18 @@
+package root.ast;
+
+import root.Token;
+
+public class LetStatement extends Statement {
```
Crain-32, line 5: Where possible, can we  maybe go into Packages with the base abstract on the top level of `ast`. Just makes it harder to read and mentally organize when you're dealing with 3-4th level `Abstract`

Crain-32 (changes requested):

Crain-32:
Apologizes for the Lack of a Response. Github didn't ping my email for one, and I've also been sick the past 2 days.

java/monkey/src/main/java/root/ast/LetStatement.java
Crain-32: I will also tack on an apologizes on this suggestion for @xNaCly , I know we had talked about avoiding nesting too many packages, but it was the only way I could think of to help with mentally with the nesting of all these abstractions. It lets me see them grouped in my IDE way better, which helps keep them "grouped" in my head at least.
Also follows normal Packaging behavior from my understanding.

java/monkey/src/main/java/root/Parser.java
RafRunner: The main reason I'm return an optional in these functions is because they can fail and I rather return an Optional (since java doesn't have a native return type) then return null of throw an Exception the the case of an error.

Suppose someone writes `if (myVar < 10 { ` forgetting to close the bracket, it would result in an error when constructing the if expression, and I think return an Optional is the best case so we don't need multiple null checks or a try - catch

java/monkey/src/main/java/root/Parser.java
RafRunner: Hmm, I don't really know how to make this function have a single return statement while keeping things readable and not repeating code, specially because the `expectPeek` has side effects. I can think of a way to do it using optionals but I don't think you'll like it:

```java
   private Optional<Statement> parseLetStatement() {
        LetStatement letStatement = new LetStatement(this.currentToken);

        return this.expectPeek(TokenType.IDENT).flatMap(identToken -> {
            letStatement.name = new Identifier(identToken, identToken.literal());

            return this.expectPeek(TokenType.ASSIGN);
        }).map((assignToken -> {

            // TODO: We're skipping the expressions until we encounter a semicolon
            while (!this.curTokenIs(TokenType.SEMI)) {
                this.nextToken();
            }

            return letStatement;
        }));
    }
```

Changing the `expectPeek` function to: 

```java
    private Optional<Token> expectPeek(TokenType type) {
        if (this.peekTokenIs(type)) {
            this.nextToken();
            return Optional.of(this.currentToken);
        }

        this.peekError(type);
        return Optional.empty();
    }
```

If you know another way to write this please let me know

java/monkey/src/main/java/root/REPL.java
RafRunner: I don't really thing a while (true) with a break is bad, it seems like a pretty common pattern to me. But I have two ideas to improve it:

```java
var scanner = new Scanner(inputStream);
var finished = false;

do {
    printStream.print(PROMPT);
    var line = scanner.nextLine();

    if (line.trim().length() != 0) {
        var lexer = new Lexer(line);

        var t = lexer.nextToken();
        while (t.type() != TokenType.EOF) {
            printStream.println(t);
            t = lexer.nextToken();
        }
    } else {
        finished = true;
    }
} while (!finished);
```
and
```java
var scanner = new Scanner(inputStream);

printStream.print(PROMPT);
var line = scanner.nextLine();

while (line.trim().length() != 0) {
    var lexer = new Lexer(line);

    var t = lexer.nextToken();
    while (t.type() != TokenType.EOF) {
        printStream.println(t);
        t = lexer.nextToken();
    }

    printStream.print(PROMPT);
    line = scanner.nextLine();
}
```
But i do believe the original one to be more clear.

java/monkey/src/main/java/root/ast/Identifier.java
RafRunner: I do believe that an Identifier should be an Expression, as it is capable of generating a value, as opposed to an let statement

image

We can see in the book that while `LetStatement` implements the `statementNode` function,`Identifier` implements the `expressionNode` function.

java/monkey/src/main/java/root/ast/Statement.java
RafRunner: I kept both functions with default empty implementations for now because in the book they haven't been implemented yet, all all implementations would just be an empty function, but I do agree on changing this when the implementation is actually needed.

java/monkey/src/main/java/root/ast/LetStatement.java
RafRunner: What about this structure?
image

java/monkey/src/main/java/root/Parser.java
RafRunner: Also, I don't really see the original implementation as having multiple return statements since the only valid return is the last one at the end. The other are merely error states that when encountered means we should stop building the Statement/Expression

Crain-32: If they are errors, we should throw Exceptions.

Status Errors can be ignored, Exceptions cannot.

java/monkey/src/main/java/root/ast/LetStatement.java
Crain-32: That Structure is better..

java/monkey/src/main/java/root/REPL.java
Crain-32: You could also make a wrapper class.
```java
public class LexerIterator implements Iterator<Token> {
    private final Lexer lexer;
    public LexerIterator(Lexer lexer) {
        this.lexer = lexer;
    }
    public static LexerIterator fromString(String input) {
       return new LexerIterator(new Lexer(input));
    }
    
    @Override
    public boolean hasNext() {
        return lexer.peek() != '\0';
    }

    @Override
    public Token next() {
        return lexer.nextToken();
    }
}
```
If you wanted to create a Token Buffer instead to match against Token Types, that would be fine. Actually might be better.

With this, we can adjust to the following
```java
printStream.print(PROMPT);
while (scanner.hasNextLine()) { // Might need some extra handling here for the initial pass, not sure
   var line = scanner.nextLine();
   LexerIterator.fromString(line).forEachRemaining(this::printToken); 
   printStream.print(PROMPT);
} // The exception would break us out of the loop.

public void printToken(Token token) {
     if (token.type() == TokenType.ILLEGAL) {
        throw new IllegalArgumentException("Illegal Token Provided", token.literal());
     }
     System.out.println(token);
}
```
Easier to read the flow of the logic, and we get the option to treat the Lexer as a forward looking Iterator.

java/monkey/src/main/java/root/ast/Identifier.java
Crain-32: In that case swap the name to  `IdentiferExpression` to more closely align with more typical Java Naming Conventions.

java/monkey/src/main/java/root/Parser.java
Crain-32: Only in the context of library creation do we need to worry about someone else. As we control the entire context, we do not need to worry about that.

Pattern Matching (something we're showcasing a bit) can also handle nulls, so I'm not worried about returning null, or even a new `EmptyStatement`

java/monkey/src/main/java/root/REPL.java
Crain-32: `while(true)` may be "common", but it's a ban common. It implies that you never leave the context of the loop, a better pattern would be this.
```java
boolean exitLoop = false;
while (!exitLoop) {
     if (exitCondition) {
        exitLoop = true;
        continue;
     }
}
```
`do/while` being more preferred.

Crain-32 (changes requested):

java/monkey/src/main/java/root/Parser.java
RafRunner: Many interpreters and compilers stop on the first error and only report on it, but I believe the idea in the book is to output all errors on a single run so the programmer can fix them all in one go, that's why I don't believe exceptions to be the best idea here (I also don't like the ideas of exceptions on general because of the hidden and disruptive control flow and all problems that come with it, and they should be only be thrown on truly unrecoverable errors or when the programmer misuses the api, but I know that's an unpopular opinion among java programmers).

I changed the implementation to a more java-esque one with no Optional. I guess my time on Rust is getting to me lol

RafRunner:
Addressed all points @Crain-32. Please mark all you fell have been addressed properly as resolved or comment further.

java/monkey/src/main/java/root/Parser.java
Crain-32: Yeah no optional and instead returning the `null` is best if you want performance.
Exceptions aren't bad (in general) as long as you don't overboard with them.

java/monkey/src/main/java/root/lexer/Lexer.java
```diff
@@ -69,7 +74,7 @@ private void advance() {
         this.pos++;
     }
 
-    private char getCc() {
+    char getCc() {
```
Crain-32, line 77: We can swap this to be `protected` at some point in the future. Since `LexerIterable` is in the same package.

java/monkey/src/main/java/root/parser/Parser.java
```diff
@@ -0,0 +1,112 @@
...
+import root.ast.expressions.IdentiferExpression;
+import root.ast.statements.LetStatement;
+import root.ast.statements.ReturnStatement;
+import root.ast.statements.Statement;
+import root.lexer.Lexer;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class Parser {
```
Crain-32, line 15: I think we could probably do something here with the `LexerIterable`, but I'm too tired this morning to come up with smart ideas.

Crain-32 (changes requested):
You have a lot of Public Properties in Classes, please swap those to private /w Getters.

java/monkey/src/main/java/root/parser/Parser.java
RafRunner: If you think about it, the Lexer itself already is an iterator (you call nextToken() until it returns EOF), the only advantage of declaring a custom Iterable is being able to use java default functions, so I don't think we would gain much from this.

java/monkey/src/main/java/root/lexer/Lexer.java
RafRunner: Changed to protected.

RafRunner:
> You have a lot of Public Properties in Classes, please swap those to private /w Getters.

I keep it that way to keep to code more boilerplate free, as if the classes where records, because I don't believe we'll ever need extra code when getting or settings values, but I'll change it since it is the java way after all

java/monkey/src/main/java/root/parser/Parser.java
Crain-32: This would be more like an `Iterable<Statement>`, but I'm thinking the right design will become more clear as we progress forward.

java/monkey/src/main/java/root/parser/Parser.java
```diff
@@ -0,0 +1,112 @@
...
+    public Program parseProgram() {
+        Program program = new Program();
+
+        while (!curTokenIs(TokenType.EOF)) {
+            try {
+                Statement statement = parseStatement();
+                if (statement != null) {
+                    program.getStatements().add(statement);
+                }
+            } catch (ParserException pe) {
```
Crain-32, line 41: Hmm. Maybe an `ErrorStatement` might fit here in the future? Then we can do something like
```java
switch (statement) {
   case ErrorStatement es -> this.errors.add(statement.es.getMessage());
   case Statement s -> program.getStatements().add(statement); // Would make more sense to have an "addStatement(Statement)" function on Program
   default -> throw IllegalStateExecption("parseStatement() should never return a value to get here");
   nextToken();
}
```
That avoid the Parser Exception, and means if this throws an Exception, we're actually dealing with an Illegal State that the program isn't designed to handle.

Once again, we can do it in another PR.

java/monkey/src/main/java/root/parser/Parser.java
```diff
@@ -0,0 +1,112 @@
...
+
+    private boolean curTokenIs(TokenType type) {
+        return this.currentToken.type() == type;
+    }
+
+    private boolean peekTokenIs(TokenType type) {
+        return this.peekToken.type() == type;
+    }
+
+    private void expectPeek(TokenType type) throws ParserException {
```
Crain-32, line 105: Given the comment about `ExceptionStatement` and such, we'd probably swap this to eventually be `private boolean expectPeek(TokenType type)`, so then we can begin doing something like this.
```java
switch (foo) {
     case LetStatement ls when expectPeek(TokenType.INDENT) -> //do something
     case LetStatement ls when !expectPeek(TokenType.INDENT) -> // do something else
}
```
Once again, just thinking about future stuff, so we can hold it off till we hit the future.

Crain-32 (approved):
Thank you for entertaining my pickiness, and for keeping up the work with the Java Portion, I hope my feedback has helped you learn some stuff as well.

@vhladko Please Merge in this PR. @RafRunner I'm assuming you'll be implementing Chapter 3 sometime this week, so keep the last comments in mind to see if we can implement some of those ideas.

java/monkey/src/main/java/root/parser/Parser.java
RafRunner: I don't know how I feel about an ErrorStatement, specially if it is a subclass of Statement, because statements in this contexts should be parts of the program the the interpreter will execute, an error caused by bad syntax is not a part of the program, it could never be added to the list of statements in the Program object or be executed, so I would strongly oppose it being a Statement, it should be it's own separate class. The bad part is that we would need to change the return types of the parsing functions to Object or something else, witch I also don't like, I honestly prefer just throwing checked exceptions in this case.

java/monkey/src/main/java/root/parser/Parser.java
RafRunner: It could be done but I don't see a reason to use switch here other than having a single return statement. We are not matching anything, foo here is always a LetStatement, if statements should be enough. I personally don't see having multiple return statements as that big of a problem in simple functions where the early returns are just error cases, I can see the point where multiple returns return valid outputs, but not for error cases.

vhladko:
Nice, let's go :)

PR #219, opened 2023-06-10
Should not use Optional as field.

andirady:
Optional should not be used as parameters & return type.


Crain-32:
I actually hadn't read up on this before. Good catch.

Crain-32 (approved):

Crain-32:
@novusnota , can you merge this in? I still don't have Write Permissions :/

PR #224, opened 2023-06-10
C# 2: Add partial parser

leo-costa:
It can currently parse `let` and `return` statements


leo-costa:
@vhladko tagging you. CODEOWNERS don't have write permissions to the repo yet correct?

LucasTavaresA:
why not have the Lexer._input as a span? you can turn it into a ref struct and remove AsSpan() from the functions

leo-costa:
> why not have the Lexer._input as a span? you can turn Lexer into a ref struct and remove AsSpan() from the functions
> 
> 

My initial version was like this when I just translated the Rust version to C#. I had no idea of how the code would be used, i.e could be called from an async method, ref struct don't work on them. Then i found out it was based on a book.

PR #225, opened 2023-06-11
Add additional `typescript-types` based lexer

AntonPieper:
For learning purposes I created my own version of the Lexer using types in typescript. I tried to write it as clearly as possible and add comments for aspiring typescript devs.


vhladko:
Hey, we really appreciate your work, but do you think that your implementation is much different from existing ones and you want to have it separate ? Just asking I'm here not for judgement :)

AntonPieper:
Originally I wrote it thinking I did everything different. But now that you ask me there are some things that are basically identical apart from names. There are some things that are different but I don't think that there needs to be another folder in this repo. Maybe I will try to incorporate some of these changes into the original code or try to optimize the compilation (in this case lexing) performance of the original code.

PR #235, opened 2023-06-13
Add Vim syntax highlighting for HolyC

xslendix:
It annoyed me Vim couldnt syntax highlight HC, so I made one to include with my code in case anyone want to look at it in a nice way.


Spongman:
You shouldn't need to `clone` in your `Dockerfile`. The `Makefile` already mounts the language-specific subdirectory at `/deez`.

PR #238, opened 2023-06-14
C#: Changed implementation to use reader instead of position counting…

TomasDurica:
… and added comments for educational purposes

### DIFF
--- a/csharp/Monkey/Parsing/Expressions.cs
+++ b/csharp/Monkey/Parsing/Expressions.cs
@@ -2,6 +2,7 @@ using Monkey.Lexing;
 
 namespace Monkey.Parsing;
 
+// This is the exact same approach we took in ../Lexing/Tokens.cs
 public static class Expressions
 {
     public static IExpression Error(string message, IToken token) => new ErrorExpression(message, token);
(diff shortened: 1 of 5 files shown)


TomasDurica:
@leo-costa - would it be too bold of me to snag your docker and makefile? Just so I will not have failing checks :)

TomasDurica:
@vhladko - don't want to bother you, but can you take a look when you will have time? :) 
Also, would you prefer to always tag you, or should I just wait :)

leo-costa:
> @leo-costa - would it be too bold of me to snag your docker and makefile? Just so I will not have failing checks :) 

Go ahead! I'm using dotnet 8 tho!

TomasDurica:
@leo-costa - many thanks :) Yes I will make sure it works, I just wanted a start point and I always feel bad when I just blatantly steal stuff :D

leo-costa:
> @leo-costa - many thanks :) Yes I will make sure it works, I just wanted a start point and I always feel bad when I just blatantly steal stuff :D 

I know you mentioned your solution doesn't really care much for speed but for what it is worth, dotnet 8 was about 23% faster than 7.

PR #243, opened 2023-06-18
feat(java): Continuing parser development up to chapter 2.6 (Infix Expressions)

RafRunner:
Implementing up to chapter 2.6 of the book and keeping pretty close to it. I posed some questions on the forms of TODOS that will need to be resolved before the merge, or we can change it later. I tried  to keep everything simple and follow conventions, and I'm open for suggestions. We should just try and keep things close to the book at least until the whole parser is finished.


RafRunner:
I'll fix the broken checks later (+-8hrs)

RafRunner:
This took a little longer than expected, but I added a Makefile and removed the cache part of the Dockerfile since in order for that to work we would need to modify the static part of the Makefile which I don't think Prime wants us to do.

Crain-32:
I'll review this tomorrow.

RafRunner:
I just realized the check got stuck because the docker image executes main and it gets stuck waiting input. 

I believe it should just run the tests.

Can't change and test it right now, I'll address it later .

java/monkey/src/main/java/root/parser/Parser.java
```diff
@@ -94,14 +95,96 @@ private Statement parseLetStatement() throws ParserException {
...
+        statement.setExpression(this.parseExpression(OperatorPrecedence.LOWEST));
+
+        if (this.peekTokenIs(TokenType.SEMI)) {
+            this.nextToken();
+        }
+
+        return statement;
+    }
+
+    private void noPrefixParseFnError(TokenType type) throws ParserException {
```
Crain-32, line 110: This function could likely just be inlined.

java/monkey/src/main/java/root/parser/Parser.java
```diff
@@ -94,14 +95,96 @@ private Statement parseLetStatement() throws ParserException {
...
+        throw new ParserException("No prefix parse function found for " + type);
+    }
+
+    private Expression parseExpression(OperatorPrecedence precedence) throws ParserException {
+        ParserSupplier<Expression> prefixFn = prefixParseFn();
+
+        if (prefixFn == null) {
```
Crain-32, line 117: Maybe swap for `Objects.requireNonNullElseThrow()`?

java/monkey/src/test/java/ParserTest.java
```diff
@@ -77,14 +68,200 @@ void testString() {
...
+
+    private void testLetStatement(Statement statement, String name) {
         Assertions.assertEquals(TokenType.LET.token().literal(), statement.tokenLiteral());
 
         if (statement instanceof LetStatement letStatement) {
             Assertions.assertEquals(name, letStatement.getName().getValue());
             Assertions.assertEquals(name, letStatement.getName().tokenLiteral());
         } else {
-            throw new AssertionError("statement not instance of LetStatement");
+            throw new AssertionError(statement.getClass().getSimpleName() + " is not instance of LetStatement");
```
Crain-32, line 264: There might be an Assertions for Classes? I know there is one for Exceptions.

java/monkey/src/main/java/root/parser/Parser.java
```diff
@@ -94,14 +95,96 @@ private Statement parseLetStatement() throws ParserException {
...
+        Expression leftExpression = prefixFn.get();
+
+        while (!peekTokenIs(TokenType.SEMI) && precedence.compareTo(peekPrecedence()) < 0) {
+            ParserFunction<Expression, Expression> infixFn = infixParseFn(peekToken.type());
+
+            if (infixFn == null) {
+                break;
+            }
+
+            nextToken();
```
Crain-32, line 129: I'm still not a fan of `nextToken()`. It feels like it should return the next token, instead of increment.
Maybe we swap to `skipNextToken(): void` and `nextToken(): Token`?

java/monkey/src/main/java/root/parser/ParserFunction.java
```diff
@@ -0,0 +1,7 @@
+package root.parser;
+
+@FunctionalInterface
+public interface ParserFunction<T, R> {
```
Crain-32, line 4: I dig the use of the `FunctionalInterface`.

Crain-32 (changes requested):
Based on your responses, I'll let it through or not. I know I was a bit picky, but we're starting to get closer to the point where choices we make now will be annoying to correct later.

java/monkey/src/test/java/ParserTest.java
RafRunner: There is one, I just wanted to cast and verify at the same time. I used the class asserting in some tests to keep nesting to a minimal.

java/monkey/src/main/java/root/parser/Parser.java
RafRunner: I kinda agree, but this is the name the author gave the function, and I believe there is some value in keeping the parallels between ours and the authors implementation.

java/monkey/src/main/java/root/parser/Parser.java
RafRunner: It does look better, but sadly Objects.requireNonNull and it's variants can only throw NullPointerException, but Iĺl inline the exception.

java/monkey/src/main/java/root/parser/Parser.java
RafRunner: Perhaps another name like `proceedToNextToken()` ? I feel like `skipNextToken()` gives the impression we're ignoring something

Crain-32: Sorry for taking a moment to respond. Yeah that works @RafRunner . After that change I'll approve.

Crain-32: @RafRunner is that a change we're okay making? Just checking around the code.

RafRunner: @Crain-32 Sorry for taking a little while, Github didnd't notify me and I was a bit busy. My next PR might be the full book completion.

Crain-32: @RafRunner just tacking this on before I approve. That works for me. Thanks for doing all this work on the Java side, I really appreciate it.

Crain-32 (approved):
@vhladko can you merge this in please? <3

Crain-32:
@vhladko Just wanted to bump the PR and remind you it's good to merge.

RafRunner:
If anyone still cares, I finished the book and merged into this PR's branch so the Java implementation would be complete.

PR #246, opened 2023-06-19
C# 2: Add identifier, integer literal, infix and prefix expression parsing and update to dotnet 8

leo-costa:


leo-costa:
@vhladko  :)

PR #254, opened 2023-06-19
Create REPL for HolyC + Update vim syntax highlight

xslendix:
This patch implements a basic line-by-line REPL. It also updates the HolyC Vim syntax highlight.


xslendix:
@vhladko can you please take a look at this PR? I want to get to working on parser and AST.

PR #258, opened 2023-06-21
Bash Makefile, improve comments, create debug.sh

ThenWhenceComethEvil:
Clarified some of the comments in the lexer.

Created initial _debug.sh_, which will eventually have more pretty
printing & debug functions as the compiler continues.

Don't think the Dockerfile is necessary. Dependencies are only
`shellcheck` and `bats`. Don't think people are so keen to check my work
as to spin up a docker image rather than installing temporarily?

### DIFF
--- a/bash-twce/Makefile
+++ b/bash-twce/Makefile
@@ -1,0 +1,14 @@
+.PHONY: help ready lint test
+
+help:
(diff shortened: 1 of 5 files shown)


ThenWhenceComethEvil:
I see there's an Action failing to run `make docker-build`. Can add one, if that would help.

PR #268, opened 2023-06-28
Feat: Complete C++ Tokenizer

orycho:
Most inmportantly this completes the Tokenizer in the cpp directory, up to the level in the "Rust lexer: Finished it" video.

Also includes several other changes:
- Keyword token types clased with C++ keywords, so capitalized all token types for uniformity
- Avoiding copy into Lexer is a good idea, but having it rely on a string_view with the contents possibly being deallocated elsewhere is potentially hazardous. Instead, move the string into the lexer.
- Conversely, tokens where overengineered using variants for literals: Made using them more dfficult, and special case for char needed only where the information is not even neeed. Replaced literals with string_view (but remember, now the literal does not live beyond the lexer's lifetime)
- Added contructor to token for easier codeing. Needed more red tape...
- replaced position based indexing into string with an iterator. Much more C++ style.
- Lexer no longer throws exception on empty input, just returns an EOF token. 
- Reduced C++ standard needed to c++20 as the only c++23 feature used was not worth it (see block inside "#if __cplusplus < 202302L" in test.cc

### DIFF
--- a/cpp/lexer/lexer.cc
+++ b/cpp/lexer/lexer.cc
@@ -1,92 +1,165 @@
 #include "lexer.hh"
+#include <cctype>
 
 
-#include <exception>
+const std::unordered_map<std::string, token_type, lexer::string_hash, std::equal_to<>> lexer::keywords {
+       {"fn",token_type::Function},
+       {"let",token_type::Let},
+       {"if",token_type::If},
+       {"else",token_type::Else},
+       {"return",token_type::Return},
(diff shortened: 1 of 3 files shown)


orycho:
Hmm, just realized the pull is based onm y master branch head, and now it also includes the parser...

orycho:
Monkey interpreter is now complete with 
* strings, arrays and hash maps
* extended set of builtin functions

PR #272, opened 2023-07-08
You can't use comments in JDSL! 😠 

uross-11:
@Crain-32
Listen here buddy, get those comments out of there before something breaks

### DIFF
--- a/jjdsl/build.xml
+++ b/jjdsl/build.xml
@@ -8,37 +8,31 @@
 
     <target name="buildJJDSL">
         <antcall target="createSource"/>
-        <!--    Base Classes   -->
         <antcall target="loadFile">
             <param name="commit" value="ade14fa068f28050615fa3960d37f6a5ab39a979"/>
             <param name="fileLocation" value="${basedir}/source/jjdsl/index.js"/>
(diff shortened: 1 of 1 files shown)


Crain-32 (approved):
I will state 3 things.

1, this is JJDSL, not JDSL >:(
2, thank you for contributing to JJDSL!
3, JJDSL is currently undergoing a full rewrite because I'm not happy with the amount of coupling between JJDSL's internals and the Interpreter. I'm almost done to the point I'm happy with it, and once I get past Chapter 3, the rewrite will be replacing this, and I'll probably release the JJDSL engine on Maven.

uross-11:
Sorry for mixing up the two, but I thought it's a good idea to remove comments out of pure respect for Tom(he is a genius btw).

> 3, JJDSL is currently undergoing a full rewrite because I'm not happy with the amount of coupling between JJDSL's internals and the Interpreter. I'm almost done to the point I'm happy with it, and once I get past Chapter 3, the rewrite will be replacing this, and I'll probably release the JJDSL engine on Maven.

I'm a measly JS dev, and probably couldn't understand for the life of me what you did in your project.

Jokes aside, I'm sure it's a nice piece of software that you are writing.

PR #278, opened 2023-10-21
(Java) fixing Dockerfile typo

RafRunner:
I accidentally typed `fac(pow(10))` instead of `puts(fac(10))` in the example program in the CMD of the Dockerfile, which just leads to an error because fac is not defined.


RafRunner:
I've included some cleanups and minor typo fixes. I'm open to suggestions.

